PR #4, opened 2019-10-07
dracut: remove systemd-tmpfiles-setup from initramfs

dongsupark:
Although no unit depends on `systemd-tmpfiles-setup.service` explicitly, it gets sometimes pulled in automatically for some reasons. As a result, the initramfs has both `systemd-tmpfiles-setup-dev.service` and `systemd-tmpfiles-setup.service`, so units like `systemd-resolved.service` cannot start because of an ordering cycle issue. That's the case especially for ARM64 architecture.

To fix that, let's remove `systemd-tmpfile-setup.service` from the initramfs.

### DIFF
--- a/dracut/99start-root/module-setup.sh
+++ b/dracut/99start-root/module-setup.sh
@@ -4,4 +4,7 @@
 
 install() {
     inst_rules "$moddir/65-start-root.rules"
+
+    rm -f "$initdir/usr/lib/systemd/system/systemd-tmpfiles-setup.service"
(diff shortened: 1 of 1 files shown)


pothos:
I wonder why there is a circle and both do not coexist well in the initramfs.

The change looks ok as it is not enabled currently.
I checked the output of `systemd-tmpfiles --cat-config` inside the initramfs to see if there is anything that we might need (excluding the `/dev` entries which are created) and some files were created by the tools itself other are missing and may not cause problems now because we don't make use of them. It is good to keep in mind that we forcefully disabled it if something breaks later because it depends on it.

pothos (approved):

PR #7, opened 2020-02-12
dracut: check both OEM IDs for flatcar and coreos

dongsupark:
For backwards compatibilities, we need to check both OEM IDs in the kernel command line, `flatcar.oem.id` and `coreos.oem.id`.

Partly addresses https://github.com/flatcar-linux/Flatcar/issues/16.

### DIFF
--- a/dracut/30ignition/ignition-generator
+++ b/dracut/30ignition/ignition-generator
@@ -42,13 +42,13 @@ if $(cmdline_bool flatcar.first_boot 0); then
     if [[ $(cmdline_arg flatcar.first_boot) = "detected" ]]; then
         add_requires ignition-quench.service
     fi
-    if [[ $(cmdline_arg flatcar.oem.id) == "packet" ]]; then
+    if [[ $(cmdline_arg flatcar.oem.id) == "packet" ]] || [[ $(cmdline_arg coreos.oem.id) == "packet" ]]; then
         add_requires flatcar-static-network.service
     fi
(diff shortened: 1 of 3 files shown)


dongsupark:
Updated it so it can simply have 2 different network files for both flatcar and coreos.

pothos (approved):

dracut/30ignition/ignition-generator
```diff
@@ -102,4 +102,9 @@ if [[ $(systemd-detect-virt || true) =~ ^(kvm|qemu)$ ]]; then
     oem_id=qemu
 fi
 
-echo "OEM_ID=$(cmdline_arg flatcar.oem.id ${oem_id})" > /run/ignition.env
+oem_cmdline="$(cmdline_arg flatcar.oem.id ${oem_id})"
+if [[ -z "${oem_cmdline}" ]]; then
```
pothos: This should be:
`if [[ "${oem_id}" = "${oem_cmdline}" ]]; then`

pothos (changes requested):

dracut/30ignition/ignition-generator
pothos: because `oem_cmdline` falls back to `oem_id`

dongsupark: Fixed

pothos (approved):

PR #8, opened 2020-02-12
Support coreos.first_boot for compatibility

pothos:
In addition to https://github.com/flatcar-linux/bootengine/pull/7 it adds support for the `first_boot` variable. In the case of GRUB it is not needed because we assume that only the image building writes the file and uses the `flatcar` path.

Edit: Added support in the udev rule, too.


dongsupark (approved):
Right, I have just realized those things. :+1:

PR #11, opened 2020-03-26
Simplify ignition-quench service and ensure 0 exit code

pothos:
The shell script propagated the error code of the [ -e coreos ]
check as exit code for the whole program because it is the last
statement that is executed.
Use the --force argument for rm and try to delete both files.
This gives a 0 exit code if the files got deleted (or never existed
but that is covered in the if condition before) but still returns
an error if the files could not be deleted.


dongsupark (approved):
Tested. Looks good. Thanks! :+1:

PR #13, opened 2020-05-22
Support the CoreOS GRUB /boot/coreos/first_boot flag file

pothos:
When a machine is migrated from CoreOS, GRUB code is not migrated and
specifies the coreos.first_boot=detected kernel parameter if the flag file
/boot/coreos/first_boot exists.
Run Ignition if the kernel parameter is present.

# How to use

See https://github.com/flatcar-linux/scripts/pull/68

# Testing done

See https://github.com/flatcar-linux/scripts/pull/68

### DIFF
--- a/dracut/30ignition/ignition-generator
+++ b/dracut/30ignition/ignition-generator
@@ -39,7 +39,7 @@ if $(cmdline_bool flatcar.first_boot 0) || $(cmdline_bool coreos.first_boot 0); 
     add_requires ignition-disks.service
     add_requires ignition-files.service
     # Only try to mount the ESP if GRUB detected a first_boot file
-    if [[ $(cmdline_arg flatcar.first_boot) = "detected" ]]; then
+    if [[ $(cmdline_arg flatcar.first_boot) = "detected" ]] || [[ $(cmdline_arg coreos.first_boot) = "detected" ]]; then
         add_requires ignition-quench.service
     fi
     if [[ $(cmdline_arg flatcar.oem.id) == "packet" ]] || [[ $(cmdline_arg coreos.oem.id) == "packet" ]]; then


dracut/30ignition/ignition-generator
```diff
@@ -39,7 +39,7 @@ if $(cmdline_bool flatcar.first_boot 0) || $(cmdline_bool coreos.first_boot 0);
     add_requires ignition-disks.service
     add_requires ignition-files.service
     # Only try to mount the ESP if GRUB detected a first_boot file
-    if [[ $(cmdline_arg flatcar.first_boot) = "detected" ]]; then
+    if [[ $(cmdline_arg flatcar.first_boot) = "detected" ]] || [[ $(cmdline_arg coreos.first_boot) = "detected" ]]; then
```
margamanterola, line 42: Could something else be setting this or is it only the grub script from the other change? If it's just the grub script, I think it would be clearer if it set only one value (flatcar.first_boot) instead of two values depending on the directory.

If there's something else that could potentially set this value, then I guess this change would be fine, but I'd like a comment explaining where the double naming comes from.

pothos: The issue is that migrated machines will only set `coreos.first_boot` always because they have the CoreOS GRUB code.

margamanterola: I wonder if we should update that GRUB code then. But anyway, in that case, please add a comment explaining that the coreos flag gets set on machines that were upgraded from coreos (and let's please NOT set it ourself).

pothos: Ah, sorry for early merging. Yes, we could add a comment, but I didn't think about it because three lines above the two cases are also covered.

pothos: Even if we updated GRUB there can still be iPXE scripts that set the coreos kernel parameters.

margamanterola: Ok, I was mostly talking about where the split came from, which is not covered by the comment above (which is talking about the conditional, not about the fact that we take into account both variables.

I think it's important to document why we are keeping both variables and **more importantly** when we intend to stop supporting both (or what would need to happen for us to stop supporting both).

PR #17, opened 2020-07-23
disk-uuid: improve logic for UUID randomization

margamanterola:
# Move UUID randomization code to its own script

We're removing the code in GRUB that detects whether the disk needs to be randomized or not (flatcar-linux/scripts#82). So, we now need to detect whether the change needs to happen during initramfs.

Before, `ignition-generator` was trying to detect whether the `disk-uuid` unit should run or not. Due to ordering and timing issues, detecting whether there's a disk with UUID 00000000-0000-0000-0000-000000000001 in `ignition-generator` doesn't work.  So, instead, move the logic to a separate script, that checks whether something needs to be done or not, and execute the unit unconditionally. The unit verifies if it needs to randomize the UUID and only does so when necessary

# How to use / Testing done

Building an image with this change plus the change in flatcar-linux/scripts#82 leads to GRUB booting successfully on a c3.medium.x86 machine followed by the disk UUID getting randomized.

WIP notice: Due to issues with current flatcar-master-alpha, I wasn't yet able to fully test this change on all platforms, I only manually tested it on Packet. Once alpha is fixed, I'll test again on all platforms.


dracut/30disk-uuid/disk-uuid.service
```diff
@@ -1,18 +1,17 @@
...
 DefaultDependencies=no
+
 Wants=local-fs-pre.target
 Before=local-fs-pre.target systemd-fsck-root.service
-Wants=systemd-udevd.service
-After=systemd-udevd.service
-Requires=%i.device
-After=%i.device
+
+Wants=systemd-udevd.service dev-disk-by\x2dlabel-ROOT.device
```
pothos: The root device can be recreated by Ignition and is not really required to exist. I suggest waiting for one of the USR partitions.

dracut/30ignition/ignition-generator
```diff
@@ -87,11 +87,8 @@ Type=oneshot
 ExecStart=/usr/sbin/ignition-setup ${nopxe:+normal} ${pxe:+pxe}
 EOF
 
-RANDOMIZE_DISK_GUID=$(cmdline_arg flatcar.randomize_disk_guid)
-if [ -n "$RANDOMIZE_DISK_GUID" ]; then
-    escaped_guid=$(systemd-escape --path "/dev/disk/by-diskuuid/$RANDOMIZE_DISK_GUID")
-    add_requires "disk-uuid@${escaped_guid}.service"
-fi
+# Unconditionally call the randomizer, it will check if it needs to act or not.
+add_requires "disk-uuid.service"
```
pothos: This should only be done if `nopxe=1`

dracut/30disk-uuid/disk-uuid.service
margamanterola: What do you mean is not required to exist? How can ROOT not exist? I guess I can try with USR-A, but given how much time I already wasted on this, I won't try until alpha test are working again.

dracut/30disk-uuid/disk-uuid.sh
```diff
@@ -0,0 +1,38 @@
...
+    echo "${DEVPATH} does not exist. Bailing out"
+    exit 1
+fi
+
+# Verify the device actually exists before trying to touch it
+DEVICE="${DEVPATH}/${RANDOMIZE_DISK_GUID}"
+if [[ -e "${DEVICE}" ]]; then
+    /usr/bin/cgpt repair ${DEVICE} && \
+    /usr/sbin/sgdisk --disk-guid=R ${DEVICE} && \
+    /usr/bin/udevadm settle
```
pothos: If the goal is to have the `/dev/disk/by-diskuuid/000…` entry go away and a new `/dev/disk/by-diskuuid/` entry to appear, it would make sense to wait for this to happen. The udevadm settle command is more like a `sleep 1` and not a reliable waiter. Maybe it's not needed and then it could be removed but if it is needed, I would prefer a real waiter.

margamanterola: I don't know what "a real waiter" would look like. But this part of the code comes from the previous unit and I'd rather not change that in this commit, since that part was working well enough already.

dracut/30disk-uuid/disk-uuid.service
pothos: The root partition can be created in Ignition and consist of a RAID or be on a different drive.

dracut/30disk-uuid/disk-uuid.sh
pothos: It was there but through the reordering it is also done at a different time which could uncover a race condition.
The waiter could first count the number of entries in `/dev/disk/by-diskuuid/`, and then wait for `/dev/disk/by-diskuuid/000…` to not exist but the number of entries be at least the same as before.

dracut/30disk-uuid/disk-uuid.service
pothos:
```suggestion
Wants=systemd-udevd.service dev-disk-by\x2dpartlabel-USR\x2dA.device
```

dracut/30disk-uuid/disk-uuid.service
```diff
@@ -1,18 +1,17 @@
...
+After=systemd-udevd.service dev-disk-by\x2dlabel-ROOT.device
```
pothos:
```suggestion
After=systemd-udevd.service dev-disk-by\x2dpartlabel-USR\x2dA.device
```

dracut/30ignition/ignition-generator
pothos:
```suggestion
if [ "${nopxe}" = 1 ]; then
    add_requires "disk-uuid.service"
fi
```

dracut/30disk-uuid/disk-uuid.sh
pothos: Let's leave it as long as it works but I think it would be cleaner to remove it and see if it still works, and if not we know the condition that we want to wait for.

dracut/30disk-uuid/disk-uuid.service
pothos: Oh, maybe this is wrong because sysroot-usr.mount maybe only works after the initramfs. Will have a look.
Edit: Updated it to point to USR-A

dracut/30disk-uuid/disk-uuid.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+
```
pothos:
```suggestion
set -e
```

dracut/30disk-uuid/disk-uuid.sh
pothos: The service failed [on Azure](http://localhost:9091/job/os/job/kola/job/azure/109/console) with exit code 1. Logs are [here](http://localhost:9091/job/os/job/kola/job/azure/109/execution/node/4/ws/_kola_temp/azure-2020-07-31-1513-48767/get_cluster_semver/jenkins-azure-25ac007432/journal.txt/*view*/)
I don't see why but I suggest to make the script more robust and report errors. I guess the `udevadm settle` command failed due to some unrelated timeout? We can either try to remove the call or log the error and just continue.

dracut/30disk-uuid/disk-uuid.sh
```diff
@@ -0,0 +1,38 @@
...
+DEVPATH="/dev/disk/by-diskuuid"
+if [[ ! -e ${DEVPATH} ]]; then
+    echo "${DEVPATH} does not exist. Bailing out"
+    exit 1
+fi
+
+# Verify the device actually exists before trying to touch it
+DEVICE="${DEVPATH}/${RANDOMIZE_DISK_GUID}"
+if [[ -e "${DEVICE}" ]]; then
+    /usr/bin/cgpt repair ${DEVICE} && \
```
pothos:
```suggestion
    /usr/bin/cgpt repair ${DEVICE}
```

dracut/30disk-uuid/disk-uuid.sh
```diff
@@ -0,0 +1,38 @@
...
+    /usr/sbin/sgdisk --disk-guid=R ${DEVICE} && \
```
pothos:
```suggestion
    /usr/sbin/sgdisk --disk-guid=R ${DEVICE}
```

dracut/30disk-uuid/disk-uuid.sh
pothos:
```suggestion
    /usr/bin/udevadm settle || echo "udevadm settle failed"
```

pothos: On all other platforms no test failed. I guess the Azure machine was very slow and also had some other problem: In the log udev is hanging for some time.

pothos: It seems that this `udevadm settle` failure on Azure is nothing new and already happens some times. Here is an output of a Azure test failing on the **main** branch – we see that `sgdisk` was running (_The operation has completed successfully_) which means that the failing process was `udevadm settle`:
```
Jul 31 12:54:38.845227 systemd[1]: Starting Generate new UUID for disk GPT dev/disk/by-diskuuid/00000000-0000-0000-0000-000000000001...
Jul 31 12:54:38.870575 cgpt[405]: Primary Header is updated.
Jul 31 12:54:38.870575 cgpt[405]: Secondary Entries is updated.
Jul 31 12:54:38.870575 cgpt[405]: Secondary Header is updated.
Jul 31 12:54:38.891154 systemd[1]: Found device Virtual_Disk OEM.
Jul 31 12:54:38.921469 systemd[1]: Found device Virtual_Disk EFI-SYSTEM.
Jul 31 12:54:38.929559 kernel: random: crng init done
Jul 31 12:54:39.018151 systemd[1]: Found device Virtual_Disk ROOT.
Jul 31 12:54:39.021250 systemd[1]: Reached target Initrd Root Device.
Jul 31 12:54:39.044066 systemd[1]: Found device Virtual_Disk USR-A.
Jul 31 12:54:40.020420 sgdisk[407]: The operation has completed successfully.
Jul 31 12:54:40.024299 kernel:  sda: sda1 sda2 sda3 sda4 sda6 sda7 sda9
Jul 31 12:55:38.791337 systemd-udevd[257]: eth0: Worker [304] processing SEQNUM=1346 is taking a long time
Jul 31 12:56:38.102290 systemd[1]: Finished dracut initqueue hook.
Jul 31 12:56:38.105000 audit[1]: SERVICE_START pid=1 uid=0 auid=4294967295 ses=4294967295 subj=kernel msg='unit=dracut-initqueue comm="systemd" exe="/usr/lib64/systemd/systemd" hostname=? addr=? terminal=? res=success'
Jul 31 12:56:38.108621 systemd[1]: Reached target Remote File Systems (Pre).
Jul 31 12:56:38.128010 kernel: audit: type=1130 audit(1596200198.105:10): pid=1 uid=0 auid=4294967295 ses=4294967295 subj=kernel msg='unit=dracut-initqueue comm="systemd" exe="/usr/lib64/systemd/systemd" hostname=? addr=? terminal=? res=success'
Jul 31 12:56:38.121946 systemd[1]: Reached target Remote File Systems.
Jul 31 12:56:38.123074 systemd[1]: Starting dracut pre-mount hook...
Jul 31 12:56:38.147712 kernel: audit: type=1130 audit(1596200198.134:11): pid=1 uid=0 auid=4294967295 ses=4294967295 subj=kernel msg='unit=dracut-pre-mount comm="systemd" exe="/usr/lib64/systemd/systemd" hostname=? addr=? terminal=? res=success'
Jul 31 12:56:38.134000 audit[1]: SERVICE_START pid=1 uid=0 auid=4294967295 ses=4294967295 subj=kernel msg='unit=dracut-pre-mount comm="systemd" exe="/usr/lib64/systemd/systemd" hostname=? addr=? terminal=? res=success'
Jul 31 12:56:38.133674 systemd[1]: Finished dracut pre-mount hook.
Jul 31 12:56:40.162062 systemd[1]: disk-uuid@dev-disk-by\x2ddiskuuid-00000000\x2d0000\x2d0000\x2d0000\x2d000000000001.service: Main process exited, code=exited, status=1/FAILURE
Jul 31 12:56:40.197052 kernel: audit: type=1130 audit(1596200200.163:12): pid=1 uid=0 auid=4294967295 ses=4294967295 subj=kernel msg='unit=disk-uuid@dev-disk-by\x2ddiskuuid-00000000\x2d0000\x2d0000\x2d0000\x2d000000000001 comm="systemd" exe="/usr/lib64/systemd/systemd" hostname=? addr=? terminal=? res=failed'
Jul 31 12:56:40.163000 audit[1]: SERVICE_START pid=1 uid=0 auid=4294967295 ses=4294967295 subj=kernel msg='unit=disk-uuid@dev-disk-by\x2ddiskuuid-00000000\x2d0000\x2d0000\x2d0000\x2d000000000001 comm="systemd" exe="/usr/lib64/systemd/systemd" hostname=? addr=? terminal=? res=failed'
Jul 31 12:56:40.162240 systemd[1]: disk-uuid@dev-disk-by\x2ddiskuuid-00000000\x2d0000\x2d0000\x2d0000\x2d000000000001.service: Failed with result 'exit-code'.
Jul 31 12:56:40.162544 systemd[1]: Failed to start Generate new UUID for disk GPT dev/disk/by-diskuuid/00000000-0000-0000-0000-000000000001.
Jul 31 12:56:40.165358 systemd[1]: Dependency failed for Initrd Default Target.
Jul 31 12:56:40.166221 systemd[1]: initrd.target: Job initrd.target/start failed with result 'dependency'.
```
Full log [here](http://localhost:9091/job/os/job/kola/job/azure/108/execution/node/4/ws/_kola_temp/azure-2020-07-31-0952-38338/coreos.ignition.security.tls/jenkins-azure-d46235a78d/journal.txt/*view*/) for the test on the **main** branch (well almost, I tested a networkd unit to exclude weave devices but this is totally unrelated).

margamanterola: Ok, I'm applying your changes. I'm not sure ignoring a udev failure is the way to go, though.

pothos (approved):
[Tests](http://localhost:9091/job/os/job/manifest/694/cldsv/) are passing on all platform with just a few provisioning failures.
Booting after installation on Packet c3.medium was also successful, yet tests didn't run because I was connected to the console which made kola think that GRUB output wasn't present.

PR #21, opened 2021-02-17
Set hostname for EC2 and OpenStack from metadata

pothos:
Similar to what is done for DigitalOcean and Packet, we can set the
hostname for EC2 and OpenStack through afterburn.
For OpenStack the name in afterburn does not match the oem_id set on
    the kernel commandline from GRUB, and a special service is added
    which sets the provider directly.

# How to use

Run kola tests and grep for `hostname` in the console/journal output.

AWS:
Compared to the current state the end result is the same because `systemd-hostnamed` sets the final hostname.
However, before that happens the hostname is currently not set while with the change here the hostname is already set as soon as the system starts.

OpenStack:
not tested

# Testing done

[AWS here](http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/2055/cldsv/)

### DIFF
--- a/dracut/30ignition/flatcar-metadata-hostname.service
+++ b/dracut/30ignition/flatcar-metadata-hostname.service
@@ -1,0 +1,13 @@
+[Unit]
+Description=Flatcar Metadata Hostname Agent
+DefaultDependencies=false
+Before=initrd.target
+After=systemd-networkd.service initrd-root-fs.target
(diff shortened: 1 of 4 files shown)


krnowak (approved):

margamanterola (approved):
LGTM, thanks. This should also fix the hostnames of AWS ARM instances which are currently not getting set, due to the polkit issue.

PR #23, opened 2021-04-07
ignition-generator: make the hostname units optional

margamanterola:
Having the hostname units as required by the initrd.target meant that if
the unit failed (for example because the network was or the metadata
service were down), the machine wouldn't start. By making it a "wants"
rather than a "requires" we allow this unit to fail without disrupting
the whole boot.

# Testing done

I ran CI with this change and it succeeded.


pothos (approved):

margamanterola:
CI passed, merging.

PR #27, opened 2021-08-02
dracut: let network-cleanup.service finish before entering the rootfs

pothos:
The network-cleanup.service failed when it wasn't fast enough to
complete before the root filesystem is switched:
  network-cleanup.service: Current command vanished from the unit file,
  execution of the command list won't be resumed.
Similar to the sysroot-boot.service we need to be finished with the
network-cleanup.service before the switch is done.

Add the ordering and conflicts directives as done in
sysroot-boot.service.

## Testing done

Done in http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/3185/cldsv/ - looks good, in the past I had seen the errors on Azure kola tests some times but of course with plain testing there is no way to know with races whether they are gone

### DIFF
--- a/dracut/03flatcar-network/network-cleanup.service
+++ b/dracut/03flatcar-network/network-cleanup.service
@@ -7,7 +7,9 @@ RefuseManualStop=true
 ConditionKernelCommandLine=!netroot
 
 PartOf=systemd-networkd.service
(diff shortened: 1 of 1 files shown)


jepio (approved):
Seems correct

PR #28, opened 2021-09-07
dracut: run emergency.target on ignition/torcx service unit failure

jepio:
# dracut: run emergency.target on ignition/torcx service unit failure

This PR isolates emergency.target when ignition units fail, ensuring that we don't end up with a boot loop in initramfs.

## How to use

```
emerge-amd64-usr bootengine coreos-kernel
```

## Testing done

Tested with invalid ignition configs from https://github.com/kinvolk/Flatcar/issues/434 and https://github.com/kinvolk/Flatcar/issues/468. In both cases the boot now stops and the user sees the ignition error on the console.

### DIFF
--- a/dracut/30ignition/ignition-disks.service
+++ b/dracut/30ignition/ignition-disks.service
@@ -16,6 +16,9 @@ After=systemd-networkd.service
 Wants=systemd-resolved.service
 After=systemd-resolved.service
 
+OnFailure=emergency.target
+OnFailureJobMode=isolate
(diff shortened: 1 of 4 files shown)


krnowak (approved):

sayanchowdhury (approved):

pothos:
Wonderful, it stops with a red error message about the wrong configuration, and shows the dracut emergency shell prompt :heart:

PR #31, opened 2021-11-03
dracut: Stop using deprecated actions in veritysetup

krnowak:
The "create" action became "open", and "remove" became "close". Also
reorder the parameters accordingly (it's a bit different for "open" vs
"create"). Also put the options before specifying the action.

CI: http://localhost:9091/job/os/job/manifest/4066/cldsv/


krnowak:
CI passed.

jepio (approved):

PR #32, opened 2021-12-20
update-bootengine: make dracut install loop driver

dongsupark:
Unlike with Kernel 5.10, dracut does not automatically install `loop.ko` with Kernel 5.15.
Explicitly install the loop module from the dracut command line.

## Testing done

CI passed for amd64: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/4434/cldsv/

- [x] Changelog entries added in the respective `changelog/` directory (user-facing change, bug fix, security fix, update)

### DIFF
--- a/update-bootengine
+++ b/update-bootengine
@@ -34,6 +34,7 @@ DRACUT_ARGS=(
     --omit multipath
     --omit network
     --add iscsi
+    --add-drivers loop
     )
 
 SETUP_MOUNTS=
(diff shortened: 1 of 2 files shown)


update-bootengine
```diff
@@ -34,6 +34,7 @@ DRACUT_ARGS=(
     --omit multipath
     --omit network
     --add iscsi
+    --drivers loop
```
pothos:
```suggestion
    --add-drivers loop
```
The help says `-d, --drivers [LIST]  Specify a space-separated list of kernel modules to exclusively include in the initramfs.` which sounds like we should take `--add-drivers [LIST]  Specify a space-separated list of kernel modules to add to the initramfs.`

dongsupark: Did so. Thanks.

tormath1 (approved):
great investigation on this, thanks !

PR #34, opened 2022-02-01
Port to Ignition v3

pothos:
- dracut/30ignition/ignition-setup.sh: migrate to Ignition v3
In Ignition v3 we don't have a "default" config anymore and there can
now be multiple "base" configurations but they have to be in a
subfolder.
Skip copying the "default" config, which should be repaced by a new OEM
systemd service that checks whether Ignition found a config or not.
Use the new "base.d" subfolder as target for the OEM "base" config.
-  dracut/99setup-root: Fix failed setup of wants dependency    
    The build log contained this error message:
      ln: failed to create symbolic link '/var/tmp/portage/sys-kernel/coreos-kernel-5.15.24/temp/dracut.Ddrb5k/initramfs//usr/lib/systemd/system/initrd.target.wants/initrd-setup-root.service': No such file or directory
    Make sure the directory exists to be able to set up the link.
- Port to Ignition v3
   The dracut/30ignition files had to be updated to match the folder
    dracut/30ignition found in the ignition repo but with some adaptions
    because Flatcar allows to overwrite the root filesystem and has a
    different initramfs service setup.
    For Ignition v3 we had to provide an implementation for is-live-image
    and ignition-kargs-helper. The kargs helper does not handle all cases
    due to the complexity of evaluation grub.cfg.

## How to use

Goes with the Ignition v3 update

Clevis is not yet supported. We need docs for the kargs support.

## Testing done
With this v3 config we have can control that the kernel arguments in `/usr/share/oem/grub.cfg` contains only the entry that should exist but no the other if it was present.
```
{
  "ignition": {
    "version": "3.3.0"
  },
  "storage": {
    "files": [
      {
        "path": "/works",
        "contents": {
          "source": "data:,helloworld%0A",
          "verification": {}
        },
        "mode": 420
      }
    ]
  },
  "kernelArguments": {
    "shouldExist": ["ishouldexist"],
    "shouldNotExist": ["ishouldnotexist"]
  }
}
```

↑ will do in coreos-overlay


tormath1 (approved):
well it looks good to me, and the CI is passing. Just have one question regarding the `Documentation=` parameter in unit files, they target CoreOS' github not sure if we should keep them or use Flatcar's URL.

pothos:
I think it's ok to use the upstream URL (not even sure if someone sees these entries). It also reduces the noise when comparing the files to upstream.

tormath1:
> It also reduces the noise when comparing the files to upstream.

Yeah, I agree too - so let's merge this. :tada:

PR #35, opened 2022-02-07
support flag file to runtime switch to cgroupv1

jepio:
# support flag file to runtime switch to cgroupv1

The idea is to detect the presence of a flag file (`/etc/flatcar-cgroupv1`), to call `systemctl switch-root` with `init.wrapper` as an argument. `init.wrapper` takes care of unmounting the currently mounted cgroup after initramfs, allowing sysroot systemd to remount cgroupv1.

The detection is done at the end of initrd-parse-etc.service, and the parameters are passed to initrd-switch-root.service through a temporary environment file.

## How to use

To be used with https://github.com/flatcar-linux/init/pull/62.

## Testing done

### DIFF
--- a/dracut/99switch-root/initrd-parse-etc-override.conf
+++ b/dracut/99switch-root/initrd-parse-etc-override.conf
@@ -1,0 +1,2 @@
+[Service]
+ExecStartPost=-/bin/sh -c "[ -f /sys/fs/cgroup/cgroup.subtree_control ] && [ -f /sysroot/etc/flatcar-cgroupv1 ] && echo INIT=/sbin/init.wrapper >/etc/switch-root.env"
--- a/dracut/99switch-root/override.conf
+++ b/dracut/99switch-root/override.conf
@@ -1,0 +1,6 @@
+[Service]
+Type=oneshot
+Environment=INIT=
+EnvironmentFile=-/etc/switch-root.env
+ExecStart=
+ExecStart=systemctl --no-block switch-root /sysroot $INIT
--- a/dracut/99switch-root/module-setup.sh
+++ b/dracut/99switch-root/module-setup.sh
@@ -11,4 +11,6 @@ install() {
         "${systemdsystemunitdir}/initrd-switch-root.service.d/override.conf"
     inst_simple "${moddir}/initrd-parse-etc-override.conf" \
         "${systemdsystemunitdir}/initrd-parse-etc.service.d/override.conf"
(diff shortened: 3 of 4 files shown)


dracut/99switch-root/initrd-parse-etc-override.conf
```diff
@@ -0,0 +1,2 @@
+[Service]
+ExecStartPost=-/bin/sh -c "[ -f /sys/fs/cgroup/cgroup.subtree_control ] && [ -f /sysroot/etc/flatcar-cgroupv1 ] && echo INIT=/sbin/init.wrapper >>/etc/switch-root.env"
```
pothos: `/etc/switch-root.env` gets created here? Then I would think that ` >/etc…` is better than appending. I also would like to have `INIT` be defined in the other case because in `overwrite.conf` it gets used in both cases.

dracut/99switch-root/override.conf
```diff
@@ -0,0 +1,5 @@
+[Service]
+Type=oneshot
+EnvironmentFile=-/etc/switch-root.env
+ExecStart=
+ExecStart=systemctl --no-block switch-root /sysroot ${INIT}
```
pothos:
```suggestion
ExecStart=systemctl --no-block switch-root /sysroot $INIT
```
see weird semantics here: https://www.freedesktop.org/software/systemd/man/systemd.service.html#Command%20lines
`$INIT` if an empty string will not be passed while `${INIT}` will be passed as empty string argument

dracut/99switch-root/initrd-parse-etc-override.conf
jepio: Yes `switch-root.env` is created here, so the ">" is better, will switch. I'm not sure about defining in both cases though, it really needs to be empty for the other cases, and that's why the EnvironmentFile directive uses `=-` to allow the file to be missing

dracut/99switch-root/override.conf
jepio: Oh, what a subtle difference. Funny that it worked either way. Will correct.

dracut/99switch-root/initrd-parse-etc-override.conf
pothos: I found no docs on the usage of undefined variables. If it works, fine, but we could also set it with `Environment="INIT="` to define it and it will then be overwritten by `EnvironmentFile` (this gets evaluated last)

pothos (approved):

PR #36, opened 2022-02-18
update-bootengine: fix containerised builds

t-lo:
dracut, when copying (installing) files into the initrd image, preserves xattrs by default. Docker containers use overlayfs, which does not support some extended atttributes.
When creating an initrd in a containerised build, dracut prints a number of warnings and errors, and the resulting initrd is subtly broken (e.g. soft-links aren't installed).

This patch detects whether update-bootengine runs in a container and if so, sets the DRACUT_NO_XATTR env variable to prevent dracut from trying to preserve xattrs while copying.

We have validated in both the containerised SDK as well as in the cork-based chroot SDK that no xattrs are used in /build/* by means of
```
    sudo find /build/amd64-usr/ -type f -exec getfattr -d {} \;
```
which came up empty.

**This change should be picked up by all active maintenance channels. The patch should be back-ported to all active maintenance branches' SDK containers.**

### DIFF
--- a/update-bootengine
+++ b/update-bootengine
@@ -104,6 +104,11 @@ else
     DRACUT_ARGS+=( "--no-kernel" )
 fi
 
+# Copying / installing files to initrd while preserving xattrs breaks dracut in certain scenarios,
+#  e.g. when run in a container.
+DRACUT_NO_XATTR=1
+export DRACUT_NO_XATTR
+
 mkdir -p "${USE_CHROOT}$(dirname "$CPIO_PATH")"
 if [[ -n "$USE_CHROOT" ]]; then
     echo "Running dracut in $USE_CHROOT"


update-bootengine
```diff
@@ -104,6 +104,13 @@ else
     DRACUT_ARGS+=( "--no-kernel" )
 fi
 
+if [ -f /.dockerenv ] ; then
+    # don't try to copy XATTR in overlayfs
+    echo "Running in a Docker container, not copying xattrs."
+    DRACUT_NO_XATTR=1
+    export DRACUT_NO_XATTR
```
tormath1: Out of curiosity, is there any difference between this and the following oneliner:
```suggestion
    export DRACUT_NO_XATTR=1
```

update-bootengine
```diff
@@ -104,6 +104,13 @@ else
     DRACUT_ARGS+=( "--no-kernel" )
 fi
 
+if [ -f /.dockerenv ] ; then
```
pothos: I think we can even make this unconditional to have a consistent behavior

t-lo: I'd rather keep it as a safety for now, to make the change minimally intrusive. I'm not sure in which other circumstances `update-bootengine` might also be used. For instance, there's support for working on `/` instead of `/build/<arch>` but I'm entirely unsure about the circumstances this might be useful for.

pothos: We are lucky it hit a test case in kola, I think it's best to have the same behavior regardless of old SDK or new SDK and whether it's building in / or /build/BOARD.

t-lo: Makes sense. Will update the PR.

update-bootengine
t-lo: I don't think there is, I just find it cleaner to first set the variable and after that export it.

pothos (approved):

PR #42, opened 2022-03-31
dracut/ignition: add setfiles to initramfs

JAORMX:
# dracut/ignition: add setfiles to initramfs

The intention is to enable ignition to relabel files.

Related Bug: https://github.com/flatcar-linux/Flatcar/issues/673

### DIFF
--- a/dracut/30ignition/module-setup.sh
+++ b/dracut/30ignition/module-setup.sh
@@ -42,7 +42,8 @@ install() {
         userdel \
         usermod \
         wipefs \
-        cryptsetup
+        cryptsetup \
+        setfiles
 
     # Flatcar: add cloud_aws_ebs_nvme_id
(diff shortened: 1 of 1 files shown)


pothos:
Thanks, when you create a draft coreos-overlay PR that uses the commit ID, we can start a test build

tormath1:
@JAORMX thanks a lot ! Do you think you could also open a PR to enable this feature on Ignition ? https://github.com/flatcar-linux/coreos-overlay/blob/main/sys-apps/ignition/ignition-9999.ebuild#L49
Then we'll update an Ignition test to assert that relabeling is done correctly on Ignition's created files.

JAORMX:
@pothos @tormath1 here it is: https://github.com/flatcar-linux/coreos-overlay/pull/1784

JAORMX:
Unfortunately, I ran out of cycles for working on this.

PR #44, opened 2022-04-28
dracut: add network kargs service

tormath1:
This service allows to read cmdline to inject network configuration as initrd kargs.

See also: https://github.com/flatcar-linux/mantle/pull/329

- [ ] Changelog entries added in the respective `changelog/` directory (user-facing change, bug fix, security fix, update) (-> coreos-overlay)

### DIFF
--- a/dracut/03flatcar-network/afterburn-network-kargs.service
+++ b/dracut/03flatcar-network/afterburn-network-kargs.service
@@ -1,0 +1,22 @@
+[Unit]
+Description=Afterburn Initrd Setup Network Kernel Arguments
+Documentation=https://coreos.github.io/afterburn/usage/initrd-network-cmdline/
+
+# This service may produce additional kargs fragments,
+# which are then consumed by dracut-cmdline(8).
+DefaultDependencies=no
+Before=dracut-cmdline.service systemd-networkd.service
+PartOf=systemd-networkd.service
+# For extra safety
+ConditionKernelCommandLine=|coreos.oem.id=vmware
+ConditionKernelCommandLine=|flatcar.oem.id=vmware
+OnFailure=emergency.target
+OnFailureJobMode=isolate
+
+[Service]
+ExecStart=/usr/bin/coreos-metadata exp rd-network-kargs --cmdline --default-value ''
+Type=oneshot
+RemainAfterExit=yes
+
+[Install]
+WantedBy=systemd-networkd.service
--- a/dracut/03flatcar-network/module-setup.sh
+++ b/dracut/03flatcar-network/module-setup.sh
@@ -10,6 +10,8 @@ depends() {
 
 # called by dracut
 install() {
+    inst_multiple coreos-metadata
+
     inst_multiple -o \
         $systemdutildir/systemd-resolved \
         $systemdsystemunitdir/systemd-resolved.service \
@@ -17,6 +19,9 @@ install() {
 
     inst_simple "$moddir/network-cleanup.service" \
         "$systemdsystemunitdir/network-cleanup.service"
+
+    inst_simple "$moddir/afterburn-network-kargs.service" \
+        "$systemdsystemunitdir/afterburn-network-kargs.service"
 
     inst_simple "$moddir/10-nodeps.conf" \
         "$systemdsystemunitdir/systemd-resolved.service.d/10-nodeps.conf"
@@ -56,4 +61,5 @@ install() {
     systemctl --root "$initdir" disable systemd-networkd.socket
 
     systemctl --root "$initdir" enable network-cleanup.service
+    systemctl --root "$initdir" enable afterburn-network-kargs.service
 }


dracut/30ignition/afterburn-network-kargs.service
```diff
@@ -0,0 +1,20 @@
...
+Before=dracut-cmdline.service
+Before=ignition-fetch.service
+After=systemd-journald.socket
+
+OnFailure=emergency.target
+OnFailureJobMode=isolate
+
+[Service]
+Environment=AFTERBURN_OPT_PROVIDER=--cmdline
+Environment=AFTERBURN_NETWORK_KARGS_DEFAULT='ip=dhcp,dhcp6'
```
pothos: Would be good to know if this has an impact if the guestinfo variable is not set or on other platforms because it would take precedence over our default networkd unit. I would rather have no default here.

tormath1: I was thinking same but it seems to be mandatory to have a default value:
* https://github.com/coreos/afterburn/blob/50ebbf2482b8c543e89b0117ef4cf8d6068fece4/src/cli/exp.rs#L63-L65 and https://github.com/coreos/afterburn/blob/50ebbf2482b8c543e89b0117ef4cf8d6068fece4/src/cli/exp.rs#L44-L47
* https://github.com/coreos/afterburn/blob/50ebbf2482b8c543e89b0117ef4cf8d6068fece4/dracut/30afterburn/afterburn-network-kargs.service#L17-L19
* I picked the value from FCOS config: https://github.com/coreos/fedora-coreos-config/blob/testing-devel/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/50-afterburn-network-kargs-default.conf#L7 (I initially found `dhcp,dhcp6` but it seems now to be `auto`)

This feature is currently only available on vmware: https://github.com/coreos/afterburn/blob/50ebbf2482b8c543e89b0117ef4cf8d6068fece4/src/initrd/mod.rs#L16-L22 so I guess it won't conflict with other providers.

pothos: It still looks like it would write the default value out to the file making the `ip=…` command line handling be triggered where it previously wasn't, or?
I think we can use an empty default meaning that the fragment file is empty and the dracut network setup would not be triggered (didn't check the details on how the fragment file is processed).

dracut/03flatcar-network/afterburn-network-kargs.service
```diff
@@ -0,0 +1,22 @@
...
+DefaultDependencies=no
+Before=dracut-cmdline.service systemd-networkd.service
+PartOf=systemd-networkd.service
+
+OnFailure=emergency.target
+OnFailureJobMode=isolate
+
+[Service]
+Environment=AFTERBURN_OPT_PROVIDER=--cmdline
+Environment=AFTERBURN_NETWORK_KARGS_DEFAULT='ip=dhcp,dhcp6'
```
pothos:
```suggestion
Environment=AFTERBURN_NETWORK_KARGS_DEFAULT=''
```
If that works we can avoid that the `/etc/systemd/network/10-dracut-cmdline-….network` file takes precedence over our default config file.

tormath1: yeah - will try that once the feature stabilized. I'm a bit afraid of hitting this: https://github.com/coreos/afterburn/blob/main/src/cli/exp.rs#L46

pothos: Reading the code it looks like an empty string is ok. The question it maybe more about how the quoting in the systemd unit here is done, which I think is correct already (cf. https://www.freedesktop.org/software/systemd/man/systemd.service.html#Command%20lines).

dracut/03flatcar-network/afterburn-network-kargs.service
```diff
@@ -0,0 +1,24 @@
...
+# For extra safety
+ConditionKernelCommandLine=|coreos.oem.id=vmware
+ConditionKernelCommandLine=|flatcar.oem.id=vmware
+OnFailure=emergency.target
+OnFailureJobMode=isolate
+
+[Service]
+Environment=AFTERBURN_OPT_PROVIDER=--cmdline
+Environment=AFTERBURN_NETWORK_KARGS_DEFAULT=''
+ExecStart=/usr/bin/coreos-metadata exp rd-network-kargs ${AFTERBURN_OPT_PROVIDER} --default-value ${AFTERBURN_NETWORK_KARGS_DEFAULT}
```
krnowak: Could you add a comment about the reason to have AFTERBURN_OPT_PROVIDER and AFTERBURN_NETWORK_KARGS_DEFAULT instead of having those values inline in ExecStart? Is this for drop-in files?

tormath1: I basically copied this service file from coreos/afterburn repository: https://github.com/coreos/afterburn/blob/main/dracut/30afterburn/afterburn-network-kargs.service and followed their way to do. 
For example, the `AFTERBURN_NETWORK_KARGS_DEFAULT` is defined later in: https://github.com/coreos/fedora-coreos-config/blob/testing-devel/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/50-afterburn-network-kargs-default.conf#L7 and this value could be platform specific.

krnowak: Yeah, so my understanding is that upstream did not define `AFTERBURN_NETWORK_KARGS_DEFAULT` in their service file, because downstream should define it through a drop-in file, just like it's done in the links you posted. Also, we are not using the service file from upstream, instead we have our own, tailored for our needs. So I'd just change it to something like:

```suggestion
ExecStart=/usr/bin/coreos-metadata exp rd-network-kargs --cmdline --default-value ''
```

This is basically the same thing. From my cursory search, neither [`AFTERBURN_NETWORK_KARGS_DEFAULT`](https://github.com/flatcar-linux/afterburn/search?q=AFTERBURN_NETWORK_KARGS_DEFAULT) nor [`AFTERBURN_OPT_PROVIDER`](https://github.com/flatcar-linux/afterburn/search?q=AFTERBURN_OPT_PROVIDER) has no special meaning within afterburn.

But if it is done with the aim of deviating from upstream in the most minimal way, then I have no problem with it staying as is.

krnowak (approved):

dracut/03flatcar-network/afterburn-network-kargs.service
tormath1: I'll go ahead and commit your suggestion - even if one wants to add a dropin to update `AFTERBURN_NETWORK_KARGS_DEFAULT` for example it will be useless since this service runs in the initramfs only.

PR #46, opened 2022-06-28
Add dracut module to persist SELinux base policy from image on boot

JAORMX:
# Add dracut module to persist SELinux base policy from image on boot

The idea is that Flatcar ships and updates the base policy, which gets
persisted (and not as a symlink) on the initramfs. Flatcar users will
then need to run a custom systemd unit to install any custom modules
needed.

### DIFF
--- a/dracut/30ignition/selinux-policy-persist
+++ b/dracut/30ignition/selinux-policy-persist
@@ -8,8 +8,8 @@
 set -euxo pipefail
 
 # Remove pre-existing policies.
-rm /etc/selinux/mcs
-rm /var/lib/selinux
+rm -rf /etc/selinux/mcs
+rm -rf /var/lib/selinux
 
 # install policies from image. Making them read/write.
 cp -a /usr/lib/selinux/mcs /etc/selinux
--- a/dracut/30ignition/selinux-policy-persist
+++ b/dracut/30ignition/selinux-policy-persist
@@ -1,0 +1,19 @@
+#!/bin/bash
+
+# Allows for auto-updating the SELinux base policy
+# while keeping the policy directory that's running
+# on the node read/write. e.g. allowing installation
+# of non-reboot-persistent custom SELinux modules.
+
+set -euxo pipefail
+
+# Remove pre-existing policies.
+rm /etc/selinux/mcs
+rm /var/lib/selinux
+
+# install policies from image. Making them read/write.
+cp -a /usr/lib/selinux/mcs /etc/selinux
+cp -a /usr/lib/selinux/policy /var/lib/selinux
+
+# Build and reload policy
+semodule -DB
--- a/dracut/30ignition/module-setup.sh
+++ b/dracut/30ignition/module-setup.sh
@@ -67,6 +67,9 @@ install() {
     # Required on system using SELinux
     inst_multiple -o setfiles
 
+    inst_script "$moddir/selinux-policy-persist" \
+        "/usr/sbin/selinux-policy-persist"
+
     inst_script "$moddir/ignition-kargs-helper" \
         "/usr/sbin/ignition-kargs-helper"
 


dracut/30ignition/selinux-policy-persist
```diff
@@ -0,0 +1,19 @@
...
+# of non-reboot-persistent custom SELinux modules.
+
+set -euxo pipefail
+
+# Remove pre-existing policies.
+rm /etc/selinux/mcs
+rm /var/lib/selinux
+
+# install policies from image. Making them read/write.
+cp -a /usr/lib/selinux/mcs /etc/selinux
```
pothos, line 15: I think this will break on update, or? The policy and the OS need to be updated together.

pothos: Understood now, this should run on every boot

dracut/30ignition/selinux-policy-persist
```diff
@@ -0,0 +1,19 @@
...
+# Allows for auto-updating the SELinux base policy
+# while keeping the policy directory that's running
+# on the node read/write. e.g. allowing installation
+# of non-reboot-persistent custom SELinux modules.
+
+set -euxo pipefail
+
+# Remove pre-existing policies.
+rm /etc/selinux/mcs
+rm /var/lib/selinux
```
pothos:
```suggestion
rm -rf /etc/selinux/mcs
rm -rf /var/lib/selinux
```
After the first boot these will be directories.

dracut/30ignition/selinux-policy-persist
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+# Allows for auto-updating the SELinux base policy
```
pothos, line 3: I think the better place for these setup steps would be in `dracut/99setup-root/initrd-setup-root`

dracut/30ignition/selinux-policy-persist
```diff
@@ -0,0 +1,19 @@
...
+# Remove pre-existing policies.
+rm /etc/selinux/mcs
+rm /var/lib/selinux
+
+# install policies from image. Making them read/write.
+cp -a /usr/lib/selinux/mcs /etc/selinux
+cp -a /usr/lib/selinux/policy /var/lib/selinux
+
+# Build and reload policy
+semodule -DB
```
pothos, line 19: I guess this needs to run in a chroot?

dracut/30ignition/selinux-policy-persist
JAORMX: could you elaborate?

dracut/30ignition/selinux-policy-persist
JAORMX: are there any samples around in these modules?

dracut/30ignition/selinux-policy-persist
pothos: Adding it there makes the order in which we prepare the rootfs more clean and I also find it better to have all steps in one place instead of having multiple modules and not knowing when which will be executed - the initramfs ordering is already complex enough.

dracut/30ignition/selinux-policy-persist
pothos: Found one:
```
dracut/35torcx/torcx-profile-populate.service
38:ExecStart=/usr/bin/chroot /sysroot /bin/bash -c "/usr/lib/coreos/torcx profile populate -v=${TORCX_VERBOSE} -n ${VERSION_ID}"
```

dracut/30ignition/selinux-policy-persist
JAORMX: Sure, I can move it there. Will this still run before ignition? That's also one requirement.

dracut/30ignition/selinux-policy-persist
JAORMX: great! I'll do that

dracut/30ignition/selinux-policy-persist
pothos: Yes, it runs before the Ignition `files` stage (where writing files to the rootfs would happen):
```
dracut/30ignition/ignition-files.service
14:Requires=initrd-setup-root.service
15:After=initrd-setup-root.service
```

PR #50, opened 2023-01-06
Manage systemd-sysext images for A/B booting

pothos:
Extensions to the generic /usr partition were in the done through a base Ignition config and extra files on the OEM partition which weren't updated. With systemd-sysext images we have a way to extend the /usr partition through an overlay and have single file updates.

Define a mechanism to discover and activate OEM sysext images and other optional Flatcar extensions that are coupled to the OS version and get updated by update-engine in an A/B scheme. The OEM sysext images are stored on the OEM partition as usual because they may be required in airgapped environments and users may want to reformat the root partition with Ignition in which case storing it on the rootfs or downloading it would not work well. If we have size limitations the inactive sysext gets moved to the rootfs but the active sysext is still placed on the OEM partition unless this also doesn't fit and users of those platforms that want to reformat the root partition would need to ensure that downloading works for them or they provide the file out of band with Ignition.
Other optional Flatcar extensions are only stored on the rootfs and are always downloaded unless a special baked image is used or the file is provided through Ignition.
This change introduces a new phase for the rootfs preparation that runs 
after Ignition. Both the phases before and after Ignition will run on   
every boot. The existing rootfs preparation had an initrd.target.wants  
link but that didn't work and is now removed.                           
The RequiresMountsFor=/sysroot/usr/share/oem directive didn't have any  
effect because there was no mount unit defined. The directive           
After=sysroot-usr.mount is already implied by                           
RequiresMountsFor=/sysroot/usr/.                                        

## Testing done
In [CI](http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/1016/cldsv/)
and with a new kola test (ran locally): https://github.com/flatcar/mantle/pull/405


dracut/10diskless-generator/diskless-generator
```diff
@@ -107,7 +107,7 @@ fi
 
 # When /usr is provided by the initrd then OEM may be also, but as tmpfs.
 # The contents are of /usr/share/oem are copied into place by 99setup-root
```
krnowak:
```suggestion
# The contents of /usr/share/oem are copied into place by 99setup-root
```

dracut/99setup-root/initrd-setup-root-after-ignition
```diff
@@ -0,0 +1,83 @@
...
+  SYSEXT_ROOT_PART="/etc/flatcar/oem-sysext/oem-${OEMID}-${VERSION}.raw"
+  SYMLINK="/sysroot/etc/extensions/oem-${OEMID}.raw"
+  ACTIVE_OEM=""
+
+  if [ -e "/sysroot/${SYSEXT_OEM_PART}" ]; then
+    ACTIVE_OEM="${SYSEXT_OEM_PART}"
+  elif [ -e "/sysroot/${SYSEXT_ROOT_PART}" ]; then
+    if [ -L "${SYMLINK}" ]; then
+      # Move the inactive sysext to the root partition to free space on the OEM partition
+      PREV_SYSEXT=$(readlink "${SYMLINK}")
```
krnowak: Is `realpath` available in initrd? You could use it, so you'll certainly get an absolute path.

dracut/99setup-root/initrd-setup-root-after-ignition
```diff
@@ -0,0 +1,83 @@
...
+    # TODO: For each OEMID, delete known old files under /usr/share/oem/ and /etc/
+    rm -f "/sysroot/usr/share/oem/sysext/migrate-oem-${OEMID}"
+  fi
+fi
+
+# Manage optional Flatcar extensions that are coupled to the OS version.
+# They are only stored on the root partition but not directly in /etc/extensions/ because there
+# can only be one file for the extension name (this could be covered by the ".v directory" proposal).
+# The enabled-sysext.conf file contains the names per line and supports comments,
+# update-engine uses this file to known which extensions to download.
```
```suggestion
# update-engine uses this file to know which extensions to download.
```

dracut/99setup-root/initrd-setup-root-after-ignition
pothos: Thanks. It's not there but with `--relative-to=/sysroot` (if I understand correctly) it would be nice to resolve the link properly. I'll test it by calling it with LD_LIBRARY_PATH from /sysroot/usr/bin/.

krnowak: There is `dracut_install grep ldconfig mountpoint systemd-tmpfiles flatcar-tmpfiles` in `module-setup.sh` in the same directory as this script. You could add `realpath` there too. It comes from coreutils, and we are already putting some coreutils stuff into initrd, so adding `realpath` shouldn't hurt.

pothos: It doesn't really help much, with `-m --relative-base=/sysroot` I get the into a similar situation as with `readlink -m` where depending on whether it's an absolute link it gets a prefix or not…

krnowak: Fair enough.

pothos: Updated it with a `sed` workaround

dracut/99setup-root/initrd-setup-root-after-ignition
```diff
@@ -0,0 +1,82 @@
...
+    if mv "/sysroot/${SYSEXT_ROOT_PART}" /sysroot/usr/share/oem/sysext/; then
+      ACTIVE_OEM="${SYSEXT_OEM_PART}"
+    else
+      echo "That failed, keeping it on root partition" >&2
+      ACTIVE_OEM="${SYSEXT_ROOT_PART}"
+    fi
+  else
+    echo "Did not find ${SYSEXT_OEM_PART} nor ${SYSEXT_ROOT_PART}" >&2
+    # TODO: Try to download it as fallback (needs starting and waiting for the network target), or maybe just give up and place a warning somewhere (motd?)
+    # URL could be https://update.release.flatcar-linux.net/${FLATCAR_BOARD}/${VERSION}/oem-${OEMID}.raw.XY
```
krnowak, line 47: Question about the 2 TODOs (not blocking this PR for me): Is this something that could be implemented inside this script, or would it need more convoluted systemd unit dependency graph to handle this case?

dracut/99setup-root/initrd-setup-root-after-ignition
```diff
@@ -0,0 +1,82 @@
...
+    mkdir -p "/sysroot/etc/extensions"
+    ln -fs "${ACTIVE_OEM}" "${SYMLINK}"
+  else
+    rm -f "${SYMLINK}"
+  fi
+
+  # Flag file created by the update-engine postinst action if both /usr partitions have a sysext and active-oem-OEMID didn't exist
+  if [ -e "/sysroot/usr/share/oem/sysext/migrate-oem-${OEMID}" ]; then
+    echo "Found migration flag, deleting known old OEM partition files" >&2
+    # TODO: For each OEMID, delete known old files under /usr/share/oem/ and /etc/
```
krnowak, line 59: Not sure I understand this TODO. Known old files that are specific to the OEM?

pothos: The files to delete are the units under `/etc/systemd/system/` that were set up by the OEM base ignition, and the files in the OEM partition itself (base ign, binaries etc).

dracut/99setup-root/initrd-setup-root-after-ignition
pothos: Yes, no other units, it would be a download with something like curl or some new binary and a signature check with gpg or openssl or some new binary.

krnowak: Right, it was a comment about waiting for the network target got me thinking, but doesn't ignition already depend on it?

krnowak (approved):
Looks good. TODOs can be addressed in follow-up work.

dracut/99setup-root/initrd-setup-root-after-ignition
pothos: Ah, yes, that's something I would do with either `systemctl start network.target` and the wait target (or directly with the wait command).

PR #52, opened 2023-01-20
Rework initrd usr mount to start Ignition/Afterburn from there

pothos:
- Rework initrd usr mount to start Ignition/Afterburn from there
    
    The size of the unified kernel image grew so much that we almost hit
    the limit of the fixed /boot partition. While we could tweak which
    kernel modules are present the largest contributors are actually
    Ignition and Afterburn (coreos-metadata). Since we also ship Afterburn
    on the /usr partition and have Ignition installed there, too (but
    currently masked), we could do what was done for Torcx and directly
    call it from there (but with LD_LIBRARY_PATH instead of chroot). There
    are a few corner cases where this was not directly possible due to the
    order of events in the initrd and these had to be reworked.
    
    This establishes a structure that allows the /sysusr/usr mount to be
    used for calling Ignition and Afterburn. This mount point was set up
    by systemd already but since we have a legacy generator, this needed
    adaption, too. The final /usr mount in /sysroot/usr is still set up.
    In the case of the systemd generator it is a bind mount, for the
    legacy generator it is a separate mount, both work but we could align
    this to also use a bind mount. The afterburn-network-kargs.service
    starts a bit later now because it depends on the /sysusr mount and
    this also means that the parse-ip-for-networkd dracut hook needed to
    run at a later stage, here solved with an own service but since the
    dracut hook mechanism with sourcing is complicated, the script still
    runs as hook to serialize the environment. The fsck check for /usr was
    also in the way and thus we write out a dummy systemd-fsck-usr.service
    since we don't need to check /usr as it's verity-protected. In the end
    with these changes we significantly reduce the unified kernel image
    size and have more breathing room for the next years. If we have to
    move some other binaries or even kernel modules that are not essential
    for loading /usr itself, we can move them out of the initrd based on
    this work. Size reduction for arm64 is 58 MB -> 52 MB, for amd64 it is
    55 MB -> 49MB.
- Align check for initrd-provided /usr
- Fix non-working norecovery option    
    The check for the norecovery option used the same condition but had
    different actions in the if branch. Also, for the systemd generator
    the unit name was wrong because it was just the unit name of the bind
    mount.
    Align the check for the norecovery option and also use the right unit
    name.

## Testing done

[CI run](http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/1137/cldsv/)

Tested the `usr=` cmdline case manually, tested the first boot `ip=10.0.2.15::10.0.2.2:255.255.255.0::eth0:off::` cmdline case (and once with `netroot=dummy`) and ran the CI to test the kargs VMware case and that Ignition works.


dracut/03flatcar-network/parse-ip-for-networkd.sh
```diff
@@ -1,17 +1,58 @@
...
+#!/bin/bash
 #
 # This script was mostly stolen from 40network/parse-ip-opts.sh.  Its
 # actions are adapted to write .network files to /etc/systemd/network
 # in the initramfs instead of using separate DHCP commands, etc.  Note
-# the bashisms.
+# the bashisms. It used to run as dracut hook, relying on afterburn to
+# process kargs for the dracut cmdline target but since that is too
+# early now that we start afterburn from /sysusr/ this script here
+# became an own systemd unit running shortly before systemd-networkd.
```
krnowak:
```suggestion
# the bashisms. It used to be run as a dracut hook, relying on afterburn to
# process kargs for the dracut cmdline target. Now that we start afterburn
# from /sysusr instead of ramdisk root, we need to mount /sysusr first. Mounting
# happens later than the dracut hook we used. Thus we added a systemd unit for
# this script. The unit runs shortly before systemd-networkd.
```

dracut/03flatcar-network/parse-ip-for-networkd.sh
```diff
@@ -1,17 +1,58 @@
...
+if [ "${APPLY-}" != "1" ]; then
+  # First run, called as dracut hook
+  {
+  # While the script only makes use of the env vars "NEEDBOOTDEV" and "netroot"
+  # we actually don't really know what env vars dracut-lib.sh and net-lib.sh
+  # will depend on in future versions and therefore, we try to preserve the
+  # environment as is.
+  for VARNAME in $(compgen -v); do
+    # Prevent leaking HEREEOF into VAL
+    [ "${VARNAME}" = "_" ] && continue
```
krnowak: Could you explain what's going on here, please? I don't get it. :)

dracut/10usr-generator/usr-generator
```diff
@@ -32,10 +32,28 @@ cmdline_arg() {
...
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=true
+EOF
+
 usr=$(cmdline_arg usr)
 usrfstype=$(cmdline_arg usrfstype auto)
 usrflags=$(cmdline_arg usrflags ro)
-if [[ "$(echo "${usrflags}" | grep -E '(^ro$|^ro,|,ro$|,ro,)')" != "" ]]; then
+if [[ "$(echo "${usrflags}" | grep -v -E '(^ro$|^ro,|,ro$|,ro,)')" != "" ]]; then
```
krnowak: Could be made simpler with:
```suggestion
if [[ "$(echo ",${usrflags}," | grep -v -F ',ro,')" != "" ]]; then
```

dracut/10usr-generator/usr-generator
```diff
@@ -61,19 +79,25 @@ esac
...
-  if [[ -z "${usr}" && -f /usr.squashfs ]]; then
+  # Here "$usr" does not contain the values of "mount.usr", hence the extra check not found in diskless-generator/ignition-generator
+  if [[ -z "${usr}" && -z "$(cmdline_arg mount.usr)" && -f /usr.squashfs ]]; then
     # Don't set "norecovery" for squashfs
     exit 0
   fi
   # Add "norecovery" to systemd-fstab-generator's unit and exit
   mount_usrflags=$(cmdline_arg mount.usrflags ro)
-  if [[ "$(echo "${mount_usrflags}" | grep -E '(^ro$|^ro,|,ro$|,ro,)')" != "" ]]; then
+  if [[ "$(echo "${mount_usrflags}" | grep -v -E '(^ro$|^ro,|,ro$|,ro,)')" != "" ]]; then
```
krnowak: Same here:
```suggestion
  if [[ "$(echo ",${mount_usrflags}," | grep -v -F ',ro,')" != "" ]]; then
```

dracut/10usr-generator/usr-generator
```diff
@@ -61,19 +79,25 @@ esac
...
     # Don't set "norecovery" when mounting rw
     exit 0
   fi
   mount_usrflags="${mount_usrflags},norecovery"
-  mkdir -p "${UNIT_DIR}/sysroot-usr.mount.d"
-  cat >"${UNIT_DIR}/sysroot-usr.mount.d/10-norecovery.conf" <<EOF
+  # sysroot-usr.mount is a bind mount from /sysusr/usr
```
krnowak, line 94: Did you document how it always was, or /sysroot/usr being a bind mount from /sysusr/usr is your change? If the latter, I can't find the change in the PR.

dracut/03flatcar-network/parse-ip-for-networkd.sh
pothos: `$_` is the last argument of the previous command and it will result in `HEREEOF` being leaked into the HEREDOC.
E.g., if you run `echo "X=\$(cat <<'HEREEOF' ; Y=$_ ; HEREEOF ; )"` multiple times in a row the first run will insert what you ran before but if you run it again and again multiple HEREEOFs will be inserted recursively.

dracut/10usr-generator/usr-generator
pothos: Thanks, good idea!

dracut/10usr-generator/usr-generator
pothos: In this branch, `sysroot-usr.mount` and `sysusr-usr.mount` come from the upstream systemd generator, yes, it's how it always was and I didn't change it (but since it's a bind mount, adding ext4-specific options doesn't make sense and we need to do it for the real mount, hence the comment).

dracut/03flatcar-network/parse-ip-for-networkd.sh
pothos: Added it as explanation

krnowak (approved):

PR #54, opened 2023-02-24
initrd-setup-root: Clean duplicate overlayfs entries

### LINKED ISSUE #647: [RFE] Handle configuration updates in /etc

pothos:
With machines having an old full /etc they would not benefit from future updates of files in /usr/share/flatcar/etc. Clean up duplicate entries to make sure that the lowerdir gets used. While this could be done in a one-time migration it is actually not bad to do this also against unwanted upcopies. The walker helper function introduced should also be used for the selective OS reset feature.

## Testing done

Booted the Alpha image and then ran:
```
ls -lah /etc
wget https://bincache.flatcar-linux.net/images/amd64/9999.9.9+kai-etc-overlay-cleanup/flatcar_test_update.gz
sudo flatcar-update -V 9999.9.9 -D -P flatcar_test_update.gz
sudo reboot
[… rebooting …]
sudo unshare -m bash -c 'umount /etc && ls -lah /etc'
```

This showed that most files/symlinks/dirs are gone and only changed files were left, e.g., `/etc/hosts`, `/etc/profile`, `/etc/iscsi` were deleted but the `/etc/systemd/system.conf` file has a different comment because the default changed and thus this file didn't get deleted.

[CI](http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/1379/cldsv/)
`shellcheck dracut/99setup-root/initrd-setup-root`
New Kola test: https://github.com/flatcar/mantle/pull/414

### DIFF
--- a/dracut/99setup-root/initrd-setup-root
+++ b/dracut/99setup-root/initrd-setup-root
@@ -1,4 +1,5 @@
-#!/bin/sh -e
+#!/bin/bash
+set -e
 # -*- mode: shell-script; indent-tabs-mode: nil; sh-basic-offset: 4; -*-
 # ex: ts=8 sw=4 sts=4 et filetype=sh
 
@@ -6,6 +7,25 @@
 # for consistency regardless if first_boot= is passed to create missing
 # files if needed and make it easier to debug. In fact, it was required
 # to use first_boot=1 with PXE to get a proper rootfs which is strange.
+
+function usrbin() {
+  local cmd="$1"
+  shift
+  LD_LIBRARY_PATH=/sysusr/usr/lib64 /sysusr/usr/bin/"${cmd}" "$@"
+}
+
+function walksysroot() {
+  local topdir="$1"
+  shift
+  local action="$1"
+  shift
+  while IFS= read -r -d '' entry; do
+    "${action}" "${entry}" || true
+  done < <(chroot /sysroot /usr/bin/find "${topdir}" -xdev -depth "$@" -print0)
+  # Do a chroot to be able to pass "-regex /etc/something" without having to prepend /sysroot
+  # in the regex. Also do the print0 as last action, after filtering.
+  true # Do not carry any last condition evaluation over as return code
+}
 
 # /etc/machine-id after a new image is created:
 COREOS_BLANK_MACHINE_ID="42000000000000000000000000000042"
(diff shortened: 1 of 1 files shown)


dracut/99setup-root/initrd-setup-root
```diff
@@ -7,6 +8,23 @@
...
+}
+
+function walksysroot() {
+  local topdir="$1"
+  shift
+  local action="$1"
+  shift
+  while IFS= read -r -d '' entry; do
+    "${action}" "${entry}" || true
+  done < <(chroot /sysroot /usr/bin/find "${topdir}" -xdev "$@" -print0 | usrbin tac -s "")
```
krnowak: Should this also use `-depth` flag? (from man: "Process each directory's contents before the directory itself.")

dracut/99setup-root/initrd-setup-root
```diff
@@ -27,6 +45,42 @@ if grep -qs "${COREOS_BLANK_MACHINE_ID}" "${MACHINE_ID_FILE}"; then
...
+  elif [ -f "${entry}" ]; then
+    # We directly compare contents, this also covers size
+    usrbin cmp -s "${entry}" "${usrentry}" && rm "${entry}" || true
+  fi
+}
+
+mkdir -p /sysroot/etc
+# Remove any files that haven't changed compared to the (new) overlay,
+# this helps that future updates will be able to take effect.
+walksysroot /etc overlaycleanup || true
```
krnowak: So my issue with this is that we may silently start updating someone's custom config that way. Let's assume that there are two versions of the config: flatcar one (in `/usr/share/flatcar/etc`) and overridden one (in `/etc`). We overrode the flatcar config, because flatcar config was using, let's say, `hash = md5`, and we wanted a different algorithm, so we use `hash= b2`. Then upstream changed their default to `hash = b2`, so the new version of Flatcar will have flatcar config with `hash = b2`. This will cause our overridden config to be deleted, because now it's the same. If later upstream changes the config again, we will pick it up after another Flatcar update (because the change could be `hash = anti-quantum-mumbo-jumbo-hashing-algo`, which we didn't want).

This example is probably very unlikely, so I would not block on it, though. Maybe I'd rather visibly document that if the admin is making changes to the configuration in /etc, it should maybe have some comment in the config, that makes it impossible to be removed by our duplicate removal (some fancy header comments or whatever).

dracut/99setup-root/initrd-setup-root
pothos: Thanks for the comment, the `tac` was exactly for that and shouldn't be needed if `-depth` is used

dracut/99setup-root/initrd-setup-root
pothos: Documentation is a good idea in general. For this case we could also make the cleanup opt-out? My main intention was to help avoid unintentional freezing which is equally bad as it makes the user carry a "modification" they are not aware of.

krnowak: Hm, maybe an opt-out at first, which will become an opt-in in half-year, for example?

krnowak: I messed up my wording. In the beginning I meant to have deduplication as an opt-in (so it's disabled by default), then half year later have it enabled by default with a way to opt-out from it. But after some discussion, maybe we can just assume that users can prepare themselves for new /etc handling through the alpha-beta-stable process, so we could just make it an enabled-by-default opt-out from the beginning.

dracut/99setup-root/initrd-setup-root
```diff
@@ -7,6 +8,24 @@
...
+  local topdir="$1"
+  shift
+  local action="$1"
+  shift
+  while IFS= read -r -d '' entry; do
+    "${action}" "${entry}" || true
+  done < <(chroot /sysroot /usr/bin/find "${topdir}" -xdev -depth "$@" -print0)
+  # Do a chroot to be able to pass "-regex /etc/something" without having to prepend /sysroot
+  # in the regex. Also do the print0 as last action, after filtering.
+}
```
krnowak: So maybe `|| true` won't be needed for the `walksysroot` invocation below.

```suggestion
  # in the regex. Also do the print0 as last action, after filtering.
  true
}
```

krnowak (approved):

PR #55, opened 2023-02-24
initrd-setup-root: Add selective OS reset handling

### LINKED ISSUE #869: [RFE] OS reset helper
## Current situation

Factory reset is something common for consumer devices but it also is useful for servers when we follow the idea of immutable infrastructure and want to "reprovision" to apply a config change but newly installing the OS takes too long or is too disruptive due to local data.
We have example usage with terraform helper scripts in the `without-instance-replacement` variants of https://github.com/flatcar/flatcar-terraform

## Impact

It's not easy for users to discard local changes. The Ignition reformat hammer is not bad but requires special knowledge and a custom ignition config entry and loses local data.

## Ideal future situation

We have a `flatcar-reset` helper that allows to either fully or selectively reset the OS.
The full reset is by wiping the rootfs (could be `rm -r`, or could be by triggering the systemd-repart factory reset mode with some special GPT disk uuid), and creates the first boot flag file for Ignition to run.
The partial reset is by doing a recommended wildcard removal before reboot (allow or deny mode, preferred is allow), then creating the first boot flag file for Ignition to run. It should preserve the machine id by writing it to grub.cfg because keeping `/etc/machine-id` is not allowed with Ignition.

pothos:
A reconfiguration of a system will lead to config drift is the local state isn't discarded. However, while this is possible with Ignition reformatting the rootfs, this big hammer is too disruptive. Introduce a way to perform a selective OS reset where a special file contains a regex for paths to preserve. This regex should be prepared by a helper tool like
  `flatcar-reset --keep-paths "/etc/ssh/" "/var/docker.*"`
that would assemble the regex to be
  `(/etc/ssh|/etc/ssh/.*|/var/docker.*)`
where the omitted slash at the end of the /etc/ssh/ is needed due to how "find" prints the paths, and the additional entry `ssh/.*` for the contents of /etc/ssh/ which would otherwise be deleted - but we should allow the user to use a final slash anyway instead of forcing the user to add `/.*` everywhere, thus the helper tool should take care of it. If the user wants to remove all contents but keep the empty dir, `^/etc/ssh` or `/etc/ssh$` or `(/etc/ssh)` or `/etc/(ssh)` would work to directly specify the wanted regex without the helper tool expanding it (see linked PR).

## How to use

Based on https://github.com/flatcar/bootengine/pull/54 - Rebase before merging

The helper tool is introduced in https://github.com/flatcar/init/pull/91

## Testing done

With the new `flatcar-reset` tool I checked that `sudo flatcar-reset --keep-machine-id --ignition-file my.ign --keep-paths '/etc/ssh/ssh_host_.*' /var/log` and a reboot will discard all state except the ssh host keys and the system logs, and will set up the `grub.cfg` cmdline for the machine ID and use the given Ignition file.

[CI](http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/1380/cldsv/)

### DIFF
--- a/dracut/99setup-root/initrd-setup-root
+++ b/dracut/99setup-root/initrd-setup-root
@@ -31,6 +31,36 @@ function walksysroot() {
 COREOS_BLANK_MACHINE_ID="42000000000000000000000000000042"
 MACHINE_ID_FILE="/sysroot/etc/machine-id"
 
+function selectiveosreset() {
+  local entry="/sysroot$1"
+  # Don't remove /sysroot itself
+  [ "${entry}" = "/sysroot" ] && return 0
+  [ "${entry}" = "/sysroot/" ] && return 0
+  # Don't remove the active /usr mount point
+  [ "${entry}" = "/sysroot/usr" ] && return 0
+  # Not really needed because find doesn't add a trailing slash but to be safe:
+  [ "${entry}" = "/sysroot/usr/" ] && return 0
+  if [ -d "${entry}" ]; then
+    # Try to delete dir, will fail if its contents are preserved
+    usrbin rmdir "${entry}" 2>/dev/null || true
+  else
+    # Delete file, Report wrong paths to nonexisting files or any other errors
+    # (journalctl -u initrd-setup-root) but don't hard fail the boot
+    rm "${entry}" || true
+  fi
+  true # Do not carry any last condition evaluation over as return code
+}
+
+# Do the selective OS reset as prepared by flatcar-reset
+if [ -s /sysroot/selective-os-reset ]; then
+  walksysroot / selectiveosreset -regextype egrep -not -regex "$(cat /sysroot/selective-os-reset)"
+  rm -f /sysroot/selective-os-reset
+  # Always remove the machine-id file because otherwise it's not a first boot.
+  # The previous value can be preserved through the systemd.machine_id=
+  # kernel parameter.
+  rm -f /sysroot/etc/machine-id
+fi
+
 # This creates the modifiable users/groups in /sysroot/etc,
 # initializing the shadow database in the process. This needs to
 # happen early, so systemd-tmpfiles can read the user info from


dracut/99setup-root/initrd-setup-root
```diff
@@ -7,10 +8,52 @@
...
+  if [ -d "${entry}" ]; then
+    # Try to delete dir, will fail if its contents are preserved
+    usrbin rmdir "${entry}" 2>/dev/null || true
+  else
+    rm "${entry}" || true # TODO: Report through flag file or fail the boot?
+  fi
+}
+
+# Do the selective OS reset as prepared by flatcar-reset
+if [ "$(cat /sysroot/selective-os-reset 2>/dev/null || true)" != "" ]; then
```
krnowak: `-s` does what you want, I think (from man: "FILE exists and has a size greater than zero")

```suggestion
if [ -s /sysroot/selective-os-reset ]; then
```

dracut/99setup-root/initrd-setup-root
```diff
@@ -7,10 +8,52 @@
...
 COREOS_BLANK_MACHINE_ID="42000000000000000000000000000042"
 MACHINE_ID_FILE="/sysroot/etc/machine-id"
 
+function selectiveosreset() {
+  local entry="/sysroot$1"
+  # Don't remove /sysroot itself
+  [ "${entry}" = "/sysroot" ] && return
+  [ "${entry}" = "/sysroot/" ] && return
+  # Don't remove the active /usr mount point
+  [ "${entry}" = "/sysroot/usr" ] && return
```
krnowak: Should we also have the check for `"/sysroot/usr/"`? Note the trailing slash.

dracut/99setup-root/initrd-setup-root
```diff
@@ -7,10 +8,52 @@
...
+    # Try to delete dir, will fail if its contents are preserved
+    usrbin rmdir "${entry}" 2>/dev/null || true
+  else
+    rm "${entry}" || true # TODO: Report through flag file or fail the boot?
+  fi
+}
+
+# Do the selective OS reset as prepared by flatcar-reset
+if [ "$(cat /sysroot/selective-os-reset 2>/dev/null || true)" != "" ]; then
+  walksysroot / selectiveosreset -regextype egrep -not -regex "$(cat /sysroot/selective-os-reset)" || true
```
krnowak: I'm probably missing something but will this remove files from /etc if needed? We are passing -xdev to find and /etc is going to be a mountpoint, so this means that /etc will be untouched (whether we want or not).

dracut/99setup-root/initrd-setup-root
pothos: Find doesn't add the slash at the end but maybe it's still a good idea

dracut/99setup-root/initrd-setup-root
pothos: `/etc` is mounted at the end, first comes the OS reset, then the overlay cleanup and both operate on the folder `/etc` (the upperdir).

dracut/99setup-root/initrd-setup-root
```diff
@@ -31,6 +31,34 @@ function walksysroot() {
...
+  [ "${entry}" = "/sysroot/" ] && return 0
+  # Don't remove the active /usr mount point
+  [ "${entry}" = "/sysroot/usr" ] && return 0
+  # Not really needed because find doesn't add a trailing slash but to be safe:
+  [ "${entry}" = "/sysroot/usr/" ] && return 0
+  if [ -d "${entry}" ]; then
+    # Try to delete dir, will fail if its contents are preserved
+    usrbin rmdir "${entry}" 2>/dev/null || true
+  else
+    rm "${entry}" || true # TODO: Report through flag file or fail the boot?
```
krnowak: I'd add `-f` so it won't fail on nonexistent files, but will still fail if there's a permission error or whatever. This is to say - we probably shouldn't be reporting anything or failing the boot when deleting nonexistent files.

```suggestion
    rm -f "${entry}" || true # TODO: Report through flag file or fail the boot?
```

krnowak (approved):
One nitpick, otherwise looks alright.

dracut/99setup-root/initrd-setup-root
pothos: If `find` hands us wrong paths that don't exist something is off and I think it's nice for it to show up in `journalctl -u initrd-setup-root`. My current take on errors would be to not have a special report file for the motd warning but rely on system logs and never fail the boot.

pothos: I've captured this in a comment now. I think failing the boot could lead to a boot loop and since failures in the initrd are also normally not easy to deal with for normal users (serial console plus the right timing for the debug session or even the GRUB menu if the initrd has a loop), I think we should always continue the boot and then it's regular debugging for the user by checking the journal.

krnowak: Sounds good. Then maybe drop the TODO and instead write that errors will be visible in the logs?

krnowak: Nevermind, you already did that.

PR #57, opened 2023-03-07
dracut/ignition: add "setfiles" binary

tormath1:
this is required to run SELinux relabelling for ignition. This should increase the initramfs size by 1M.

<hr>

Originally sent by @JAORMX

### DIFF
--- a/dracut/30ignition/module-setup.sh
+++ b/dracut/30ignition/module-setup.sh
@@ -130,4 +130,6 @@ install() {
         "/usr/bin/coreos-metadata"
     inst_script "$moddir/ignition-wrapper" \
         "/usr/bin/ignition"
+    inst_script "$moddir/setfiles-wrapper" \
+        "/usr/bin/setfiles"
 }
--- a/dracut/30ignition/setfiles-wrapper
+++ b/dracut/30ignition/setfiles-wrapper
@@ -1,0 +1,2 @@
+#!/bin/sh
+LD_LIBRARY_PATH=/sysusr/usr/lib64 exec /sysusr/usr/sbin/setfiles "$@"


JAORMX (approved):

dracut/30ignition/module-setup.sh
```diff
@@ -41,7 +41,8 @@ install() {
         userdel \
         usermod \
         wipefs \
-        cryptsetup
+        cryptsetup \
+        setfiles
```
pothos: The new approach against an ever-growing initramfs is to add a `setfiles-wrapper` file instead:
https://github.com/flatcar/bootengine/blob/flatcar-master/dracut/30ignition/coreos-metadata-wrapper
and install the wrapper file as `setfiles`:
https://github.com/flatcar/bootengine/blob/flatcar-master/dracut/30ignition/module-setup.sh#L127

tormath1: Oh good - thanks for the input.

JAORMX (approved):

dracut/30ignition/module-setup.sh
```diff
@@ -130,4 +130,6 @@ install() {
         "/usr/bin/coreos-metadata"
     inst_script "$moddir/ignition-wrapper" \
         "/usr/bin/ignition"
+    inst_script "$moddir/ignition-wrapper" \
```
pothos:
```suggestion
    inst_script "$moddir/setfiles-wrapper" \
```

pothos: (In the future if we add more of them we could also change the wrapper to be used with symlinks and evaluate `$0`)

pothos (approved):

PR #58, opened 2023-03-17
Prevent the OEM mount point from showing up before Ignition mounts

pothos:
Ignition may run a manual mount command for the OEM partition if the
    user specifies it. This fails if the OEM mount unit is triggered early,
    because it can't be mounted twice, thus we need to prevent that from
    happening by unmounting the OEM partition (but not the tmpfs OEM mount
    for PXE-provided OEM contents). To spare users from even dealing with
    the OEM partition specification in Ignition we can ensure it is mounted
    before Ignition writes files.

## Testing done
Ran https://github.com/flatcar/mantle/pull/421 manually for the image from http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/1516/cldsv/


dracut/30ignition/ignition-files.service
```diff
@@ -27,4 +30,6 @@ Before=initrd-parse-etc.service
 Type=oneshot
 RemainAfterExit=yes
 EnvironmentFile=/run/ignition.env
+# Flatcar: Make sure that the OEM mount point is there even if it shortly was away
+ExecStartPre=-systemctl start sysroot-usr-share-oem.mount
```
krnowak, line 34: Is `RequiresMountsFor` not enough here?

pothos: Not if `ignition-mount.service` doesn't mount it again after the `umount` pre-start workaround because it doesn't have it defined in the config. The `RequiresMountsFor` gets evaluated in parallel will the other services starting up, so the mount unit is present and fulfills the requirement but before this service here is actually started it may be gone.

krnowak: Blows my mind. :)

I suppose it's something that can't be made simpler because of backward-compatibility.

krnowak (approved):

dracut/30ignition/ignition-files.service
pothos: We could solve it with more downstream patches in Ignition but that's not the direction I wanted to take :upside_down_face:

pothos:
Beta (3510) is also affected, I'll look into a backport before it goes Stable.

Edit: Done, also had one additional single word change backported to ensure the ordering for initrd-setup-root when Ignition doesn't run.

PR #59, opened 2023-03-27
dracut/99setup-root: change owner of /run/log/journal

tormath1:
Give the correct ownernship otherwise the tmpfiles configuration will not apply because of mismatching machine-id - which makes "journalctl --user" to fail

### DIFF
--- a/dracut/99start-root/module-setup.sh
+++ b/dracut/99start-root/module-setup.sh
@@ -4,7 +4,4 @@
 
 install() {
     inst_rules "$moddir/65-start-root.rules"
-
-    rm -f "$initdir/usr/lib/systemd/system/systemd-tmpfiles-setup.service"
-    rm -f "$initdir/usr/lib/systemd/system/sysinit.target.wants/systemd-tmpfiles-setup.service"
 }
--- a/dracut/99start-root/module-setup.sh
+++ b/dracut/99start-root/module-setup.sh
@@ -4,4 +4,6 @@
 
 install() {
     inst_rules "$moddir/65-start-root.rules"
+
+    sed -i 's/After=local-fs.target /After=/' "$initdir/usr/lib/systemd/system/systemd-tmpfiles-setup.service"
 }


dracut/99setup-root/initrd-setup-root
```diff
@@ -86,6 +86,10 @@ systemd-tmpfiles --root=/sysroot --create \
     baselayout.conf baselayout-usr.conf \
     baselayout-home.conf
 
+# Give the correct ownernship otherwise the tmpfiles configuration will not
+# apply because of mismatching machine-id - which makes "journalctl --user" to fail.
+chown --recursive root:systemd-journal /run/log/journal/
```
pothos: Wait, this reminds me of https://github.com/flatcar/Flatcar/issues/799
I think the problem could be that `systemd-tmpfiles` doesn't find the `systemd-journal` group because it's not part of `/sysroot/etc/group` but only loaded via the nss-altfiles plugin from our database in baselayout.
I think as long as we don't change systemd-tmpfiles to somehow resolve the nss stuff in the given chroot, we should all `systemd-journal` as special group to copy like `docker`: https://github.com/flatcar/baselayout/blob/flatcar-master/scripts/flatcar-tmpfiles#L13
This will also be a workaround for the linked issue (at least for the journal but not for `kvm` - maybe add that, too? At some point we should fix the tools, though).

pothos: Based on the assumption that this existing tmpfiles rule here would set the group:
```
/usr/lib/tmpfiles.d/systemd.conf:z /run/log/journal 2755 root systemd-journal - -
/usr/lib/tmpfiles.d/systemd.conf:Z /run/log/journal/%m ~2750 root systemd-journal - -
```

tormath1: Yeah I checked this initially and the issue is not to change the owner in `/sysroot/run/log/journal` but in `/run/log/journal` (the one in the initramfs) so I tried with `systemd-tmpfiles --create systemd.conf` but in the end, I thought it would be overkill so just changed the ownership of the `/run/log/journal`.

pothos: Ah, ok, right, `/sysroot/run` isn't mounted and this setup would take place later and not in the call we do here.
If this is about `/run/log/journal` of the initrd I think it's at the wrong place. Edit: Maybe `99journald-conf`?

pothos: Oh, found this here: https://github.com/flatcar/bootengine/commit/b0b7416d91659988d99e2c0c3ba7e5897d4a0e7d

pothos: I think it's worth trying to enable it again and see if the ordering cycle is still there and how to resolve this. Moving things closer to upstream behavior can avoid other future problems.

tormath1: CI is not over but we can see failures on ARM64: http://jenkins.infra.kinvolk.io:8080/job/container/job/test/8859/console - will try to investigate on it.

pothos: I think we can try `sed -i 's/After=local-fs.target/After=local-fs-pre.target/g' "$initdir/usr/lib/systemd/system/systemd-tmpfiles-setup.service"`

pothos: I tried it out manually with `rd.shell rd.break=cmdline` (`=pre-mount` or `=pre-pivot` would be too late) on a qemu arm64 VM.
After the VM booted I see this:
```
core@localhost ~ $ ls -alh /run/log/
total 0
drwxr-xr-x.  3 root root             60 Mar 28 03:10 .
drwxr-xr-x. 30 root root            780 Mar 28 03:16 ..
drwxr-sr-x.  4 root systemd-journal  80 Mar 28 03:15 journal
core@localhost ~ $ ls -alh /run/log/journal/
total 0
drwxr-sr-x. 4 root systemd-journal 80 Mar 28 03:15 .
drwxr-xr-x. 3 root root            60 Mar 28 03:10 ..
drwxr-s---. 2 root systemd-journal 60 Mar 28 03:10 931d6b0b22a24775bbd156cfaf0e98f4
drwxr-s---. 2 root systemd-journal 60 Mar 28 03:15 bd9ce2b134744c6db1aca45238fd9848
core@localhost ~ $ ls -alh /run/log/journal/*
/run/log/journal/931d6b0b22a24775bbd156cfaf0e98f4:
total 2.5M
drwxr-s---. 2 root systemd-journal   60 Mar 28 03:10 .
drwxr-sr-x. 4 root systemd-journal   80 Mar 28 03:15 ..
-rw-r-----. 1 root systemd-journal 2.5M Mar 28 03:14 system.journal

/run/log/journal/bd9ce2b134744c6db1aca45238fd9848:
total 2.5M
drwxr-s---. 2 root systemd-journal   60 Mar 28 03:15 .
drwxr-sr-x. 4 root systemd-journal   80 Mar 28 03:15 ..
-rw-r-----. 1 root systemd-journal 2.5M Mar 28 03:16 system.journal
core@localhost ~ $ ls -alh /run/log/journal/*/*
-rw-r-----. 1 root systemd-journal 2.5M Mar 28 03:14 /run/log/journal/931d6b0b22a24775bbd156cfaf0e98f4/system.journal
-rw-r-----. 1 root systemd-journal 2.5M Mar 28 03:16 /run/log/journal/bd9ce2b134744c6db1aca45238fd9848/system.journal
```

pothos: (I have no answer for why this happens only on arm64 and why the upstream file wants `local-fs` - probably more for the real system and not really needed in the initrd, but no idea how to encode this conditionally)

tormath1: CI is done. AMD64 is failing too for the very same reason. Ignition's run seems to be the point of failure (as it pass the `update` test)
```
Mar 27 19:56:35.796020 systemd[1]: ignition-fetch.service: Found ordering cycle on systemd-resolved.service/start
Mar 27 19:56:35.796028 systemd[1]: ignition-fetch.service: Found dependency on systemd-tmpfiles-setup.service/start
Mar 27 19:56:35.796036 systemd[1]: ignition-fetch.service: Found dependency on local-fs.target/start
Mar 27 19:56:35.796043 systemd[1]: ignition-fetch.service: Found dependency on local-fs-pre.target/start
Mar 27 19:56:35.796051 systemd[1]: ignition-fetch.service: Found dependency on ignition-disks.service/start
Mar 27 19:56:35.796058 systemd[1]: ignition-fetch.service: Found dependency on ignition-fetch.service/start
``` 

It looks like we differ from coreos/ignition on this:
```
# Flatcar:
Wants=systemd-networkd.service
After=systemd-networkd.service
Wants=systemd-resolved.service
After=systemd-resolved.service
Requires=local-fs-pre.target
Before=local-fs-pre.target
```

(https://github.com/coreos/ignition/blob/main/dracut/30ignition/ignition-fetch.service & https://github.com/flatcar/bootengine/blob/3ba6216024c75d502b9696d05f4f02594d52efbb/dracut/30ignition/ignition-fetch.service#L22-L32)

Do you remember why we have this dependency:
```
 Requires=local-fs-pre.target 
 Before=local-fs-pre.target 
```
?

pothos: It's needed because Ignition can repartition the rootfs (https://github.com/flatcar/bootengine/blob/flatcar-master/dracut/30ignition/ignition-disks.service#L24)

dracut/99start-root/module-setup.sh
```diff
@@ -4,7 +4,4 @@
 
 install() {
     inst_rules "$moddir/65-start-root.rules"
```
pothos, line 6:
```suggestion
    inst_rules "$moddir/65-start-root.rules"
    
    sed -i 's/After=local-fs.target/After=local-fs-pre.target/g' "$initdir/usr/lib/systemd/system/systemd-tmpfiles-setup.service"
```

dracut/99setup-root/initrd-setup-root
pothos: As said, I think we can tweak the service to run a bit earlier, too. I've moved the comment into a suggestion.

dracut/99start-root/module-setup.sh
tormath1: Mm
```
network.target: Found ordering cycle on systemd-resolved.service/start
network.target: Found dependency on systemd-tmpfiles-setup.service/start
network.target: Found dependency on local-fs-pre.target/start
network.target: Found dependency on ignition-kargs.service/start
network.target: Found dependency on ignition-fetch.service/start
network.target: Found dependency on network.target/start
network.target: Job systemd-resolved.service/start deleted to break ordering cycle starting with network.target/start
```

So we dropped the dependency on `local-fs.target` but still have this `local-fs-pre` dependency

pothos: Seems my manual tests didn't cover this case. Ok, I think we can even drop the `local-fs-pre` - I don't think we depend on a certain filesystem mount that would need to be done - `/run` will be there already.

pothos (approved):

dracut/99start-root/module-setup.sh
```diff
@@ -5,6 +5,5 @@
...
 
-    rm -f "$initdir/usr/lib/systemd/system/systemd-tmpfiles-setup.service"
-    rm -f "$initdir/usr/lib/systemd/system/sysinit.target.wants/systemd-tmpfiles-setup.service"
+    sed -i 's/After=local-fs.target /After=/' "$initdir/usr/lib/systemd/system/systemd-tmpfiles-setup.service"
```
pothos, line 8: A bit more future proof would be something like that to remove `local-fs.target` and any space before and after it, so that it doesn't directly have to be after the `After=` with an element behind it but one also has to avoid removing it if it is the only occurrence in the line because `After=` would reset the dependencies specified before, or? For now I think it's fine and when it breaks it will be easy to adjust.

PR #60, opened 2023-03-30
initrd-setup-root: Set up more /var directories from packages

pothos:
The base_image_var.conf contents are generated during image build and
    don't originally come from tmpfiles rules but from package
    installation. For those rules coming from a package's tmpfiles I think
    t's enough to let them do the job at their regular boot stage but for
    those not from tmpfiles we don't really know whether this works for
    them, so I added it as precaution. Also, add info on where
    /var/log/journal gets set up.

## How to use

Backport to Alpha

## Testing done

[Done](http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/738/cldsv/)

### DIFF
--- a/dracut/99setup-root/initrd-setup-root
+++ b/dracut/99setup-root/initrd-setup-root
@@ -82,9 +82,15 @@ done
 /usr/sbin/flatcar-tmpfiles /sysroot
 
 # Initialize base filesystem without /etc
+# Don't run all tmpfiles rules but only the essential ones
+# required to let systemd do its work correctly
+# (baselayout.conf must contain /var/log/journal for it to
+# be created early enough, base_image_var.conf has other
+# directories that packages installed). The rest will be
+# done later as usual through systemd-tmpfiles-setup.service.
 systemd-tmpfiles --root=/sysroot --create \
     baselayout.conf baselayout-usr.conf \
-    baselayout-home.conf
+    baselayout-home.conf base_image_var.conf
 
 # Remove our phony id. systemd will initialize this during boot.
 if grep -qs "${COREOS_BLANK_MACHINE_ID}" "${MACHINE_ID_FILE}"; then


dracut/99setup-root/initrd-setup-root
```diff
@@ -82,9 +82,13 @@ done
...
 
 # Initialize base filesystem without /etc
+# Don't run all tmpfiles rules but only the essential ones
+# required to let systemd do its work correctly. The rest
+# will be done later as usual.
 systemd-tmpfiles --root=/sysroot --create \
     baselayout.conf baselayout-usr.conf \
-    baselayout-home.conf
+    baselayout-home.conf base_image_var.conf \
+    var.conf
```
krnowak: Where the `base_image_var.conf` and `var.conf` come from?

pothos: All these files are under `/sysroot/usr/lib/tmpfiles.d/`

tormath1: I got an error from the initramfs when trying to apply this conf:
```
                bash-5.1# systemd-tmpfiles --root=/sysroot --create var.conf
                /sysroot/usr/lib/tmpfiles.d/var.conf:15: Failed to resolve group 'utmp'.
```

krnowak: Sorry for the confusion. I meant that I know that baselayout-home.conf comes from our baselayout project, but I don't know from which project base_image_var.conf and var.conf come from. I couldn't find them in systemd, init or baselayout. That's why I'm asking.

pothos: `base_image_var.conf` is generated based on package contents, `var.conf` comes from systemd.

Ok, sounds like we should not add `var.conf` directly but only copy the wanted line somehow

krnowak: > Ok, sounds like we should not add `var.conf` directly but only copy the wanted line somehow

flatcar-tmpfiles in baselayout could also copy the utmp group to solve the issue Mathieu had.

tormath1:
```suggestion
systemd-tmpfiles --root=/sysroot --create var.conf --exclude-prefix /var/log/wtmp --exclude-prefix /var/log/btmp --exclude-prefix /var/log/lastlog
```

pothos: Ah, interesting, I went with https://github.com/flatcar/baselayout/pull/29 to only work on Flatcar-specific things and be independent from systemd changes

tormath1:
Note: it's possible to quickly identify if the system is persisting logs by running: `journactl --verify`: if you only see `/run/log/journal/*` it means the issue is still present. 
I'm looking to provide tests to avoid future regressions.

pothos:
Looks ok now:
```
journalctl --verify
PASS: /run/log/journal/c7ff69fdb5844e33a22092259bad1823/system.journal                                                                             
PASS: /var/log/journal/6716a3388bb147ec955254cbe893efc6/system.journal
```
After the reboot I also see the first boot:
```
journalctl --list-boots
IDX BOOT ID                          FIRST ENTRY                 LAST ENTRY                 
 -1 272c012b16a04d55b5c6d67f1ce8c85e Mon 2023-04-03 12:47:37 UTC Mon 2023-04-03 12:50:49 UTC
  0 e953f09c5c494cf2a61dc1c13f27f2a6 Mon 2023-04-03 12:50:54 UTC Mon 2023-04-03 12:51:12 UTC
```

dracut/99setup-root/initrd-setup-root
```diff
@@ -82,9 +82,12 @@ done
 /usr/sbin/flatcar-tmpfiles /sysroot
 
 # Initialize base filesystem without /etc
+# Don't run all tmpfiles rules but only the essential ones
+# required to let systemd do its work correctly. The rest
+# will be done later as usual.
 systemd-tmpfiles --root=/sysroot --create \
     baselayout.conf baselayout-usr.conf \
-    baselayout-home.conf
+    baselayout-home.conf base_image_var.conf
```
krnowak, line 93: Since you are adding `/var/log` and `/var/log/journal` in `baselayout.conf`, is adding `base_image_var.conf` here necessarry?

pothos: It's not necessary for the topic at hand, it contains things like `/var/log/audit` and `/var/db/sudo`. I thought it's good to set things up at this point to prevent similar problems showing up again. The `base_image_var.conf` contents are generated during image build and don't originally come from tmpfiles rules but from package installation. For those rules coming from a package's tmpfiles I think it's enough to let them do the job at their regular boot stage but for those not from tmpfiles we don't really know whether this works for them, so I added it as precaution.

pothos: Have updated the commit message.

krnowak: Fair enough, thanks for the updated comment.

krnowak (approved):

tormath1 (approved):

PR #61, opened 2023-04-04
/usr/share/oem -> /oem

krnowak:
CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/1784/cldsv


krnowak:
Reworded the commit.

dongsupark (approved):
Looks good.

Though AFAIU we should merge this PR when https://github.com/flatcar/mantle/pull/423 is ready as well.

krnowak:
> Looks good.
> 
> Though AFAIU we should merge this PR when [flatcar/mantle#423](https://github.com/flatcar/mantle/pull/423) is ready as well.

Thanks for the review. And yes, it will be merged only when all four PRs are ready. (mantle, scripts, init and bootengine)

dracut/99setup-root/initrd-setup-root
```diff
@@ -153,6 +153,6 @@ mkdir -p /sysroot/.etc-work
 mount -t overlay overlay -o lowerdir=/sysroot/usr/share/flatcar/etc,upperdir=/sysroot/etc,workdir=/sysroot/.etc-work,redirect_dir=on,metacopy=off,noatime /sysroot/etc
 
 # PXE initrds may provide OEM
-if [ -d /usr/share/oem ] && mountpoint --quiet /sysroot/usr/share/oem; then
-    cp -Ra /usr/share/oem/. /sysroot/usr/share/oem
+if [ -d /oem ] && mountpoint --quiet /sysroot/oem; then
+    cp -Ra /oem/. /sysroot/oem
```
pothos: This will break existing custom initrds - not sure if that is used but since we don't know, it needs to copy from `/usr/share/oem/.`

pothos:
```suggestion
if [ -d /usr/share/oem ] && mountpoint --quiet /sysroot/oem; then
    cp -Ra /usr/share/oem/. /sysroot/oem
```

pothos:
The mount unit should take care of creating the OEM mount point folder if it does not exist but when it's not activated during boot (I think that shouldn't be the case but who knows) it would be nice to ensure that it still exists, or?
How about defining it like `/boot` in https://github.com/flatcar/baselayout/blob/flatcar-master/tmpfiles.d/baselayout.conf#L1?

dracut/99setup-root/initrd-setup-root
krnowak: Reverted the change.

EDIT: Partially, that is. We are still copying things to `/sysroot/oem`.

krnowak:
> The mount unit should take care of creating the OEM mount point folder if it does not exist but when it's not activated during boot (I think that shouldn't be the case but who knows) it would be nice to ensure that it still exists, or? How about defining it like `/boot` in https://github.com/flatcar/baselayout/blob/flatcar-master/tmpfiles.d/baselayout.conf#L1?

Good idea, please see https://github.com/flatcar/baselayout/pull/30

pothos (approved):

sayanchowdhury (approved):

PR #62, opened 2023-05-31
setup-root: Detect initial sysext MVP

pothos:
- setup-root: Detect initial sysext MVP
    
    We have implemented the building of OEM sysext images but the update
    path is still in progress. Instead of blocking the migration for new
    deployments, we will lose the version coupling by using "initial" as
    version in the file name and SYSEXT_LEVEL instead of VERSION_ID for the
    extension release file.
    Detect the initial sysext MVP path to activate the OEM sysext.
- setup-root: Read old OEM files from the migration flag file
    
    The list of old files per OEM needs to be maintained somewhere. Instead
    of doing this in the initrd script, we can use the update-engine
    post-inst action which already is a place for migration logic.
    Expect that the post-inst action writes all old OEM files to clean up to
    the migration flag file that triggers the deletion.

## How to use

Requires changes in https://github.com/flatcar/scripts/pull/858/files#diff-3b2def625e0988b47d34144d412cb632e946cd13b7b57ad3918b4322cbef9262R165 to use SYSEXT_LEVEL and in https://github.com/flatcar/scripts/pull/858/files#diff-1390361e20f8ec710779da5680655da8aef20d972df05b5c0c4ca5e3ad80a1adR552 to overwrite the version to be `initial`

## Testing done

to be done with the scripts PR

### DIFF
--- a/dracut/99setup-root/initrd-setup-root-after-ignition
+++ b/dracut/99setup-root/initrd-setup-root-after-ignition
@@ -13,6 +13,7 @@ OEMID=$({ grep -m 1 -o "^ID=.*" /sysroot/oem/oem-release || true ; } | cut -d = 
 
 # The active-oem-OEMID file gets created by the update-engine postinst action if both old and new /usr partitions have a sysext
 if [ "${OEMID}" != "" ] && [ -e "/sysroot/oem/sysext/active-oem-${OEMID}" ]; then
+  INITIAL_MVP="/oem/sysext/oem-${OEMID}-initial.raw"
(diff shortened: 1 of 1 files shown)


krnowak:
I'll pull this PR into https://github.com/flatcar/scripts/pull/858.

krnowak (approved):

pothos:
I'll create a follow up for the download now that we know the file name for the fallback download from release artifacts (`oem-azure.raw` without compression).

PR #64, opened 2023-06-12
setup-root: Download sysext images if required

### LINKED ISSUE #1073: OEM sysext: Fallback download
## Current situation

The initial MVP for the OEM sysext image got merged, currently only Azure is migrated and the sysext image won't be updated and therefore is independent from the OS version (but still does dynamic linking and could break).
In the future the sysext image will be updated.

## Impact

- Case A (now): When the inital MVP sysext image breaks and we don't have updates, users won't have working OEM tools and would have to redownload manually.
- Case B (future): When the OEM sysext image is expected to be updated but not provided in the update payload, it will be missing.

## Ideal future situation

- Case A: As simpler workaround they could manually delete the old sysext image file and reboot for the fitting file to be downloaded.
- Case B: A fallback download is in place to fetch the missing OEM sysext image. Since this could happen also through deletion it would be best to have this done in the initrd and available on every boot.

## Implementation options

For case B the fallback download could also be done in the update-engine post-inst action if, e.g., Nebraska is configured wrongly and only provides the generic update payload. However, I think it's enough to have this fallback in the initrd for now as it covers more cases.

pothos:
The sysext images required for the OEM or optional extensions should be downloaded if not found. For OEM sysext images they can be missing if the user deleted them or if the additional update payload didn't wasn't used (e.g., a custom Nebraska server or flatcar-update without the right OEM payload). For optional extensions it is expected that they get downloaded on-demand on first boot (like with the OEM sysext we also need a fallback)
Add download logic for OEM sysext images as fallback method. Also use the same download logic for optional extensions where they get downloaded on first-boot (or on reboot as fallback when the update tool didn't provide the right payload).

## Testing done

see https://github.com/flatcar/scripts/pull/906

### DIFF
--- a/dracut/99setup-root/initrd-setup-root-after-ignition
+++ b/dracut/99setup-root/initrd-setup-root-after-ignition
@@ -1,12 +1,70 @@
 #!/bin/bash
 set -euo pipefail
 
+function usrbin() {
+  local cmd="$1"
+  shift
+  LD_LIBRARY_PATH=/sysusr/usr/lib64 /sysusr/usr/bin/"${cmd}" "$@"
+}
+
+function usrcurl() {
+  usrbin curl -fsSL --retry-delay 1 --retry 60 --retry-connrefused --retry-max-time 60 --connect-timeout 20 "$@"
+}
+
+function download_and_verify() {
+  # Downloads release artifact to /sysroot/$name and verifies $name.sig with gpg
+  # Expects the env vars: FLATCAR_BOARD, VERSION
+  local name="$1"
+  local channel=""
+  case $(echo "${VERSION}" | cut -d . -f 2) in
+    0) channel="alpha" ;;
+    1) channel="beta" ;;
+    2) channel="stable" ;;
+    3) channel="lts" ;;
+    *) ;;
+  esac
+  local URLS=()
+  if [ "${channel}" != "" ]; then
+    URLS+=("https://${channel}.release.flatcar-linux.net/${FLATCAR_BOARD}/${VERSION}/${name}")
+  fi
+  URLS+=("https://bincache.flatcar-linux.net/images/${FLATCAR_BOARD/-usr}/${VERSION}/${name}")
+  local URL=""
+  for URL in "${URLS[@]}"; do
+    usrcurl -o "/sysroot/${name}" "${URL}" || { rm -f "/sysroot/${name}" ; continue ; }
+    usrcurl -o "/sysroot/${name}.sig" "${URL}.sig" || { rm -f "/sysroot/${name}.sig" ; continue ; }
+    break
+  done
+  if [ ! -e "/sysroot/${name}" ] || [ ! -e "/sysroot/${name}.sig" ]; then
+    echo "Failed to download required sysext image ${name}" >&2
+    exit 1 # Fail the boot
+  fi
+  local GPG_KEY=""
+  local GPG_LONG_ID=""
+  # Extract public key from flatcar-install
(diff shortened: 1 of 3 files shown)


dracut/99setup-root/initrd-setup-root-after-ignition
```diff
@@ -1,12 +1,62 @@
...
+}
+
+function download_and_verify() {
+  # Downloads release artifact to /sysroot/$name and verifies $name.sig with gpg
+  # Expects the env vars: FLATCAR_BOARD, VERSION
+  local name="$1"
+  local URLS=()
+  URLS=(
+    # Not quoted, will generate four entries
+    $(printf "https://%s.release.flatcar-linux.net/${FLATCAR_BOARD}/${VERSION}/${name}\n" alpha beta stable lts)
```
krnowak: Could we reduce the number of URLs by mapping the second number in VERSION to channel name? (0 - alpha, 1 - beta, 2 - stable, 3 - lts)

dracut/99setup-root/initrd-setup-root-after-ignition
```diff
@@ -49,9 +99,30 @@ if [ "${OEMID}" != "" ] && [ -e "/sysroot/oem/sysext/active-oem-${OEMID}" ]; the
...
   else
-    echo "Did not find ${SYSEXT_OEM_PART} nor ${SYSEXT_ROOT_PART}" >&2
-    # TODO: Try to download it as fallback (needs starting and waiting for the network target), or maybe just give up and place a warning somewhere (motd?)
-    # URL could be https://update.release.flatcar-linux.net/${FLATCAR_BOARD}/${VERSION}/oem-${OEMID}.raw.XY
+    echo "Did not find ${SYSEXT_OEM_PART} nor ${SYSEXT_ROOT_PART}, downloading" >&2
+    systemctl start --quiet systemd-networkd systemd-resolved
+    download_and_verify "oem-${OEMID}.raw"
+    mkdir -p /run/_oem
+    mount "/sysroot/oem-${OEMID}.raw" /run/_oem/
+    if grep -q SYSEXT_LEVEL=1.0 /run/_oem/usr/lib/extension-release.d/extension-release.oem-qemu ; then
```
```suggestion
    if grep -q SYSEXT_LEVEL=1.0 /run/_oem/usr/lib/extension-release.d/extension-release.oem-${OEMID} ; then
```

dracut/99setup-root/initrd-setup-root-after-ignition
pothos: Thanks, good idea!

dracut/99setup-root/initrd-setup-root-after-ignition
```diff
@@ -1,12 +1,69 @@
...
+  local GPG_LONG_ID=""
+  # Extract public key from flatcar-install
+  GPG_KEY=$(tr '\n' '_'  < /sysroot/usr/bin/flatcar-install | grep -Po 'GPG_KEY="\K.*?(?=")' | tr '_' '\n')
+  GPG_LONG_ID=$(grep -Po '^GPG_LONG_ID="\K.*?(?=")' /sysroot/usr/bin/flatcar-install)
+  export GNUPGHOME=/run/_gpg
+  mkdir -p "${GNUPGHOME}"
+  usrbin chmod 700 "${GNUPGHOME}"
+  usrbin gpg --batch --quiet --import <<< "${GPG_KEY}"
+  if ! usrbin gpg --batch --trusted-key "${GPG_LONG_ID}" --verify "/sysroot/${name}.sig" "/sysroot/${name}"; then
+    rm -f "/sysroot/${name}.sig" "/sysroot/${name}"
```
krnowak, line 51: Also `rm -rf "${GNUPGHOME}"`.

krnowak (approved):

PR #68, opened 2023-07-26
initrd-setup-root: Improve migration for OEM systemd-sysext images

pothos:
Combined with https://github.com/flatcar/update_engine/pull/24 these changes improve the migration to OEM systemd-sysext images and the handling of extra Flatcar extensions.

## Testing done

Tested with update-engine PR

### DIFF
--- a/dracut/99setup-root/initrd-setup-root-after-ignition
+++ b/dracut/99setup-root/initrd-setup-root-after-ignition
@@ -156,14 +156,14 @@ if [ "${OEMID}" != "" ] && [ -e "/sysroot/oem/sysext/active-oem-${OEMID}" ]; the
   if [ -e "/sysroot/oem/sysext/migrate-oem-${OEMID}" ]; then
     echo "Found migration flag, deleting known old OEM partition files" >&2
     # For each OEMID, delete known old files under /oem/ and /etc/ based on the contents of the flag file
-    # (The list is maintained in the update-engine post-inst action)
+    # (The list is maintained in the coreos-base/misc-files package)
     while IFS="" read -r entry; do
       if [ "${entry}" = "/etc/systemd/system/multi-user.target.wants/oem-cloudinit.service" ] && [ -L "/sysroot/etc/systemd/system/multi-user.target.wants/oem-cloudinit.service" ]; then
         ln -fs /usr/lib/systemd/system/oem-cloudinit.service /sysroot/etc/systemd/system/multi-user.target.wants/oem-cloudinit.service || true
       elif [ "${entry}" != "" ]; then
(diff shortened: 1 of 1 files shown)


dracut/99setup-root/initrd-setup-root-after-ignition
```diff
@@ -165,9 +171,12 @@ fi
 # Manage optional Flatcar extensions that are coupled to the OS version.
 # They are only stored on the root partition but not directly in /etc/extensions/ because there
 # can only be one file for the extension name (this could be covered by the ".v directory" proposal).
-# The enabled-sysext.conf file contains the names per line and supports comments,
-# update-engine uses this file to know which extensions to download.
-for NAME in $(grep -o '^[^#]*' /sysroot/etc/flatcar/enabled-sysext.conf || true); do
+# The enabled-sysext.conf file is read from /etc and /usr and contains one name per line,
+# and when the name is prefixed with a "-" it means that the extension should be disabled if enabled by default in the file from /usr.
+# It may contain comments starting with "#" at the beginning of a line or after a name.
```
krnowak, line 176: I wouldn't bother with allowing comments after the name, it only complicates things. Unless the file format is not something we have a control over.

dracut/99setup-root/initrd-setup-root-after-ignition
```diff
@@ -165,9 +171,12 @@ fi
...
+# The file is also used in update-engine to know which extensions to download.
+# Note that we don't need "{ grep || true ; }" to suppress the match return code because in for _ in $(grep...) return codes are ignored
+for NAME in $(grep -h -o '^[^#]*' /sysroot/etc/flatcar/enabled-sysext.conf /sysroot/usr/share/flatcar/enabled-sysext.conf | grep -v -x -f <(grep '^-' /sysroot/etc/flatcar/enabled-sysext.conf | cut -d - -f 2-) | grep -v -P '^(-).*'); do
```
krnowak, line 179: If we don't allow the comments after the name, then this could be simpler, like I suggested in [the init PR](https://github.com/flatcar/init/pull/101#discussion_r1279165816).

pothos: That would simplify `'^[^#]*'` but not the rest, I think it's good to follow the https://en.wikipedia.org/wiki/Robustness_principle

krnowak: That is probably a good principle to follow, in principle. :) So let's keep it like this.

But personally, for cases where I am the sole producer and consumer of some protocol/file/whatever, I go strict/conservative in both ways.

krnowak (approved):

PR #69, opened 2023-07-28
dracut: depend on net-lib module

LaszloGombos:
# [Discuss dracut upstream change]

Somewhat abusing the PR process a bit I wanted to reach out from the dracut upstream team to get feedback on an upstream PR that we hope would benefit this project as well. 

Upstream dracut is considering to introduce a new dracut module called net-lib that would manage installing net-lib.sh (so that bootengine does not need to do that "manually) 

https://github.com/dracutdevs/dracut/pull/2250

If upstream lands, that would be only be release around Dec this year or later, so plenty of time to coordinate this work.

Please feel free to comment directly on the upstream bug (and if desired close this PR)

### DIFF
--- a/dracut/03flatcar-network/module-setup.sh
+++ b/dracut/03flatcar-network/module-setup.sh
@@ -5,7 +5,7 @@
 
 # called by dracut
 depends() {
-    echo systemd-networkd
+    echo net-lib systemd-networkd
 }
 
 # called by dracut
@@ -41,10 +41,6 @@ install() {
 
     inst_simple "$moddir/zz-default.network" \
         "$systemdutildir/network/zz-default.network"
-
-    # install net-lib.sh regardless of its parent module's status
-    inst_simple "$moddir/../40network/net-lib.sh" /lib/net-lib.sh ||
-    dfatal 'Could not install net-lib.sh from the network module'
 
     # add a hook to generate networkd configuration from ip= arguments
     inst_hook cmdline 99 "$moddir/parse-ip-for-networkd.sh"


pothos:
Thanks for reaching out, good to know that when updating

LaszloGombos:
@pothos would you say that this is a welcomed/good change from your project perspective ?

pothos:
No real opinion, I think we only use the `ip_to_var` function of it in our `parse-ip-for-networkd.sh` script

LaszloGombos:
@pothos The more I learn about Flatcar Linux the more I think this is a great and fascinating way to deploy Linux.

I noticed that https://github.com/flatcar/scripts/tree/main/sdk_container/src/third_party/coreos-overlay/sys-kernel/dracut seem to point to a somewhat old version of dracut. Is this version in use for Flatcar. If in use, can I help in any way to update this version ?

I also did noticed that https://github.com/flatcar-hub/gentoo/tree/master/sys-kernel/dracut point to an up-to-date version of dracut, but I was not able to comprehend how the very different versions of dracut actually used.

Please let me know if there is a better way to engage with the project than this issue.

pothos:
> I noticed that https://github.com/flatcar/scripts/tree/main/sdk_container/src/third_party/coreos-overlay/sys-kernel/dracut seem to point to a somewhat old version of dracut. Is this version in use for Flatcar. If in use, can I help in any way to update this version ?

Thanks! Yes, that is the file that needs updating following [these steps](https://github.com/flatcar/scripts/tree/main/sdk_container/src/third_party/coreos-overlay#updating). For PRs there is a CI run that we have to trigger - it helps if you have a [local build env](https://www.flatcar.org/docs/latest/reference/developer-guides/sdk-modifying-flatcar/) but it's not a hard requirement.

The other gentoo repo is for upstream PRs.

PR #71, opened 2023-09-20
initrd-setup-root-after-ignition: Use update key for initrd sysext DL

pothos:
The OEM and "official" Flatcar systemd-sysext images have a download mechanism in the initrd. For OEM extensions this is only a fallback in case the user deleted the files because now we have all things in place for the update to ensure that OEM payloads are there. The enabling of "official" Flatcar extensions on provisioning still would primarily go through this mechanism and therefore, it should not use the image signing key but the update key to have a single way of verifying the extensions.
Switch the extension download to use the update server and the update key for verification of the payloads.

## Testing done

[Kola test `cl.sysext.fallbackdownload` still passes](http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/2544/cldsv/)

### DIFF
--- a/dracut/99setup-root/initrd-setup-root-after-ignition
+++ b/dracut/99setup-root/initrd-setup-root-after-ignition
@@ -12,60 +12,30 @@ function usrcurl() {
 }
 
 function download_and_verify() {
-  # Downloads release artifact to /sysroot/$name and verifies $name.sig with gpg
+  # Extracts release artifact to /sysroot/$final_name
   # Expects the env vars: FLATCAR_BOARD, VERSION
-  local name="$1"
-  # TODO: We should use update.release.flatcar-linux.net and then decode the payload
-  local channel=""
-  case $(echo "${VERSION}" | cut -d . -f 2) in
-    0) channel="alpha" ;;
-    1) channel="beta" ;;
-    2) channel="stable" ;;
-    3) channel="lts" ;;
-    *) ;;
-  esac
-  local URLS=()
-  if [ "${channel}" != "" ]; then
-    URLS+=("https://${channel}.release.flatcar-linux.net/${FLATCAR_BOARD}/${VERSION}/${name}")
+  local final_name="$1"
+  local name="${final_name/.raw/.gz}"
+  URL="https://update.release.flatcar-linux.net/${FLATCAR_BOARD}/${VERSION}/${name}"
+  # Check for scripts:sdk_container/src/third_party/coreos-overlay/coreos-base/coreos-au-key/files/developer-v1.pub.pem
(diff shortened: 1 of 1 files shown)


dracut/99setup-root/initrd-setup-root-after-ignition
```diff
@@ -12,60 +12,29 @@ function usrcurl() {
...
-    *) ;;
-  esac
-  local URLS=()
-  if [ "${channel}" != "" ]; then
-    URLS+=("https://${channel}.release.flatcar-linux.net/${FLATCAR_BOARD}/${VERSION}/${name}")
+  local final_name="$1"
+  local name="${final_name/.raw/.gz}"
+  URL="https://update.release.flatcar-linux.net/${FLATCAR_BOARD}/${VERSION}/${name}"
+  if [ "$(usrbin md5sum /sysroot/usr/share/update_engine/update-payload-key.pub.pem | cut -d " " -f 1)" = "7192addf4a7f890c0057d21653eff2ea" ]; then
+    URL="https://bincache.flatcar-linux.net/images/${FLATCAR_BOARD/-usr}/${VERSION}/flatcar_test_update-${name}"
```
krnowak, line 22: Could you add a comment about this magic number (a hash of some testing key?)

dracut/99setup-root/initrd-setup-root-after-ignition
```diff
@@ -12,60 +12,29 @@ function usrcurl() {
...
-    rm -f "/sysroot/${name}.sig" "/sysroot/${name}"
-    rm -rf "${GNUPGHOME}"
-    echo "Failed to verify required sysext image ${name}" >&2
-    exit 1 # Fail the boot
-  fi
-  rm "/sysroot/${name}.sig"
-  rm -rf "${GNUPGHOME}"
+  rm -f "/sysroot/${name}"
+  usrcurl -o "/sysroot/${name}" "${URL}" || { rm -f "/sysroot/${name}" ; return 1 ; }
+  PROTOPATH=/usr/share/update_engine/ usrbin unshare -m sh -c "mount --rbind /dev /sysroot/dev/ && mount -t proc /proc /sysroot/proc/ && chroot /sysroot /usr/share/update_engine/decode_payload /usr/share/update_engine/update-payload-key.pub.pem \"/${name}\" \"/${final_name}\"" || { rm -f "/sysroot/${final_name}" ; echo "Failing boot" >&2 ; exit 1; }
```
krnowak: Should maybe the PROTOPATH env var be specified somewhere inside the `sh -c` parameter? It probably works where it is now, but it's quite far away from the actual user of the variable.

dracut/99setup-root/initrd-setup-root-after-ignition
pothos: Yes, that's the dev key, done.

dracut/99setup-root/initrd-setup-root-after-ignition
pothos: Done

krnowak (approved):

PR #72, opened 2023-09-21
ignition: support 'ami' OEM ID.

tormath1:
See: https://github.com/flatcar/scripts/pull/1083#issuecomment-1729780975


pothos (commented):
Also needs changes in
```
  63   │     if [[ $(cmdline_arg flatcar.oem.id) == "ec2" ]] || [[ $(cmdline_arg coreos.oem.id) == "ec2" ]]; then
```

PR #73, opened 2023-10-02
ignition: Add partprobe to initrd

pothos:
The sgdisk tool does not update the kernel partition table in contrast to other similar tools. Often udev can detect the changes but not always as experienced when adding a new partition on Flatcar's boot disk. Instead of implicitly relying on some other component to re-read the kernel partition table, trigger the re-read with partprobe. This is proposed in https://github.com/coreos/ignition/pull/1717

## Testing done

Done in https://github.com/flatcar/scripts/pull/1202

### DIFF
--- a/dracut/30ignition/module-setup.sh
+++ b/dracut/30ignition/module-setup.sh
@@ -37,6 +37,7 @@ install() {
         mkswap \
         nvme \
         sgdisk \
+        partprobe \
         useradd \
         userdel \
         usermod \


dracut/30ignition/module-setup.sh
```diff
@@ -37,6 +37,7 @@ install() {
         mkswap \
         nvme \
         sgdisk \
+        partprobe \
```
tormath1, line 40: Should we continue to add things directly into the initramfs? My understanding was to now use wrappers: https://github.com/flatcar/bootengine/commit/06199d5d8f791e3b54748178e24644b8a9121e61

pothos: I wanted to have it consistent and the increase wasn't too much in the end.

tormath1 (approved):

dracut/30ignition/module-setup.sh
pothos: Some symlink trick to a multi call bash script would be nice to make the wrappers shorter. The initrd size is also bloated by the firmware that we don't really need in there - we actually would only need the firmware for the disk and from there on could load the other firmware (network etc.).

PR #76, opened 2023-10-13
initrd-setup-root-after-ignition: Clean up incomplete move target

pothos:
When possible we prefer to have the active OEM systemd-sysext live on the OEM partition to support cleaning up the rootfs without bricking the instance. However, "mv" doesn't clean up the incomplete target file which would fill the OEM disk, and that could cause future problems. Clean up the incomplete move target that filled the OEM disk when trying to store the active OEM sysext image there.

## How to use

Backport to Alpha/Beta because it is a potential bug fix.

## Testing done

in https://github.com/flatcar/scripts/pull/1272


krnowak (approved):

krnowak:
Let's get this in for the releases this week.

PR #77, opened 2023-11-06
Remove torcx: Torcx was removed from the base image

pothos:
Torcx was removed from the base image because we now use systemd-sysext for swapping out Docker/containerd.
Remove the Torcx steps from the initrd.

## How to use

Check that booting doesn't fail anymore when `/etc/torcx/next-profile` exists.

## Testing done

It boots successful even with the `/etc/torcx/next-profile` file present.


t-lo (approved):
CI Tests are all green: https://github.com/flatcar/scripts/pull/1344#issuecomment-1796678143
LGTM, thank you Kai.

PR #79, opened 2023-11-29
initrd-setup-root: Clean up whiteout entries for files to be recreated

pothos:
The removal of files in the overlay present in the lowerdir creates whiteout entries that mask the lowerdir entries. For those files that have a tmpfile rule for creation, a reboot would cause the file to be created in the upperdir, meaning this file is not updated from the lowerdir when it changes. In addition we have filtered out some tmpfile rules that caused upcopies (symlinks and directories) which meant that removing the /etc/resolv.conf symlink didn't bring it back after reboot. To make files from the lowerdir show up if they have a tmpfile rule that normally would recreate them we keep a list of whiteout entries that we clean up on boot. This also prevents freezing files because systemd-tmpfiles does not need to recreate them in the upperdir.

## How to use

Backport to Stable

## Testing done

Removal of `/etc/resolv.conf` and reboot brings it back.

Note that not all files have a tmpfiles rule, e.g., `sshd_config` has not. We could identify files like that which need to exist for a working system by running `rm -rf /etc/*` and rebooting.
This is different from the removal of `/etc` contents after `umount /etc` which works without a problem and brings all files back.

### DIFF
--- a/dracut/99setup-root/initrd-setup-root
+++ b/dracut/99setup-root/initrd-setup-root
@@ -66,6 +66,17 @@ if [ -s /sysroot/selective-os-reset ]; then
   rm -f /sysroot/etc/machine-id
 fi
 
+# Remove any user-created whiteouts for files that have a tmpfiles
+# rule which normally would recreate them (we use the lowerdir for that).
+while IFS="" read -r entry ; do
+  entry="/sysroot${entry}"
+  # The -c check for character devs also guards against empty strings and nonexisting files
+  # The stat command prints the major and minor device type in decimal
+  if [ -c "${entry}" ] && [ "$(stat --printf='%Hr %Lr\n' "${entry}")" = "0 0" ]; then
+    rm "${entry}" || true
+  fi
+done < /sysroot/usr/share/flatcar/etc-no-whiteouts
+
 # This creates the modifiable users/groups in /sysroot/etc,
 # initializing the shadow database in the process. This needs to
 # happen early, so systemd-tmpfiles can read the user info from


dracut/99setup-root/initrd-setup-root
```diff
@@ -66,6 +66,16 @@ if [ -s /sysroot/selective-os-reset ]; then
   rm -f /sysroot/etc/machine-id
 fi
 
+# Remove any user-created whiteouts for files that have a tmpfiles
+# rule which normally would recreate them (we use the lowerdir for that).
+while IFS="" read -r entry ; do
+  entry="/sysroot${entry}"
+  # The -c check for character devs also guards against empty strings and nonexisting files
+  if [ -c "${entry}" ] && [ "$(stat --printf='%Hr %Lr\n' "${entry}")" = "0 0" ]; then
```
krnowak, line 75: Please write a comment what these Hr and Lr format sequences are (on my Ubuntu 22.04, `man stat` does not show them).

krnowak (approved):

dracut/99setup-root/initrd-setup-root
pothos: Interesting, it's in the Fedora man page:
```
       %Hr    major device type in decimal, for character/block device special files

       %Lr    minor device type in decimal, for character/block device special files
```

krnowak: Yeah, I just have old coreutils (8.32), whereas Fedora and Flatcar have 9.3.

pothos:
Tested with https://github.com/flatcar/mantle/pull/488

dracut/99setup-root/initrd-setup-root
pothos: Oh, will add a comment

PR #80, opened 2023-12-18
ignition: Remove partprobe dependency and use partx

pothos:
Partx is already present, but not explicitly mentioned. We can use it instead of partprobe which is not needed anymore after upstreaming our Ignition patch (https://github.com/flatcar/scripts/pull/1403).

## How to use

With new PR in https://github.com/flatcar/scripts/pull/3161

## Testing done

### DIFF
--- a/dracut/30ignition/module-setup.sh
+++ b/dracut/30ignition/module-setup.sh
@@ -37,7 +37,7 @@ install() {
         mkswap \
         nvme \
         sgdisk \
-        partprobe \
+        partx \
         useradd \
         userdel \
         usermod \


chewi (approved):
Nice. I checked a recent initrd. The only partprobe reference is an informational message in sgdisk.

bootengine might largely become redundant soon if my plan goes ahead, but it's not quite there yet.

pothos:
The scripts PR build pipeline couldn't access the git ref here: https://github.com/flatcar/scripts/pull/3161 Is there something else I need to do with `EGIT_COMMIT` compared to the CROS thing we used before? Can it build from branch commit refs?

chewi:
You need to set `EGIT_BRANCH`.

PR #84, opened 2024-01-22
generators: Don't write files other than unit files

pothos:
Systemd now executes generators in a sandbox to make sure they only generate units and don't have other side effects:
https://github.com/systemd/systemd/blob/166bcaf95fbdd036f99705d2204138c36eb95836/NEWS#L1583

Port our generator usage to comply with the new behavior so that we can
update our systemd version. Move the Ignition preparations into the
ignition-setup script, setup the fsck overwrite unit at install time,
and remove the mkdir calls because Mount units will do this when needed.

## Testing done

[see scripts PR](https://github.com/flatcar/scripts/pull/1598)


dracut/10usr-generator/module-setup.sh
```diff
@@ -12,4 +12,18 @@ install() {
...
+    # which prevents (the systemd-generated) sysusr-usr.mount from starting before Ignition
+    # but we need it because we want to start Ignition from /sysusr/usr/ already.
+   cat > "$initdir/etc/systemd/system/systemd-fsck-usr.service" <<EOF
+# Set up by 10usr-generator/module-setup.sh
+[Unit]
+DefaultDependencies=no
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=true
```
krnowak, line 26: I'm wondering if this should be kept more-or-less in sync with systemd - upstream seems to have more stuff in the unit.

krnowak: Or maybe we should carry a patch to systemd, so no overwriting needs to be done here?

pothos: This is a minimal unit, to have it start instead of masking it. What we would need upstream is a way of resetting the After= dependencies through a drop-in…

krnowak (approved):

PR #85, opened 2024-02-09
Added flatcar systemd-tmpfiles-setup-dev-early.service

ader1990:
systemd-tmpfiles-setup-dev-early.service is required to be added instead of systemd-tmpfiles-setup-dev.service, as the kmod-static-nodes service needs to run before the systemd-tmpfiles-setup-dev service.

Otherwise, the static links like /dev/vhost-net, will not be created.

Once systemd >= 253 is used, the systemd-tmpfiles-setup-dev-early.service definition can be removed, as it is already built in.

`--gracefuly` flag is not supported for the systemd 252 version, so it had to be removed.

See systemd upstream definition:
https://github.com/systemd/systemd/blob/main/units/systemd-tmpfiles-setup-dev-early.service

## Testing done


ader1990:
While testing this PR, an issue was seen with oem-cloudinit service failing. Fixed by https://github.com/flatcar/bootengine/pull/86.

pothos (approved):

pothos:
You can merge (and then update the scripts PR to use the merge commit ID)

PR #87, opened 2024-02-19
add dracut i18n_module required for systemd

ader1990:
add dracut i18n module required for systemd-vconsole-setup.

### DIFF
--- a/update-bootengine
+++ b/update-bootengine
@@ -29,11 +29,11 @@ DRACUT_ARGS=(
     --force
     --no-hostonly
     --no-compress
-    --omit i18n
     --omit lvm
     --omit multipath
     --omit network
     --add iscsi
+    --add i18n
     --add-drivers "loop brd drbd nbd rbd mmc_block xen-blkfront zram libarc4 lru_cache zsmalloc"
     )
 


pothos:
To reduce size we could remove the `/usr/share/unimaps/`, `/usr/share/consolefonts/` and `/usr/share/keymaps/` folders

pothos:
> To reduce size we could remove the `/usr/share/unimaps/`, `/usr/share/consolefonts/` and `/usr/share/keymaps/` folders

The best way is adding a Gentoo package install mask for these paths (needs to be done for the package that introduces them to the base image).

ader1990:
> > To reduce size we could remove the `/usr/share/unimaps/`, `/usr/share/consolefonts/` and `/usr/share/keymaps/` folders
> 
> The best way is adding a Gentoo package install mask for these paths (needs to be done for the package that introduces them to the base image).

will try it out asap, thanks.

ader1990:
> > To reduce size we could remove the `/usr/share/unimaps/`, `/usr/share/consolefonts/` and `/usr/share/keymaps/` folders
> 
> The best way is adding a Gentoo package install mask for these paths (needs to be done for the package that introduces them to the base image).

Just to be sure on what we expect to be the reduced size, I took a snapshot of those 3 dir sizes:

```
1.3M    ./rootfs-1/usr/share/keymaps/
368K    ./rootfs-1/usr/share/unimaps/
960K    ./rootfs-1/usr/share/consolefonts/
```

ader1990:
> To reduce size we could remove the `/usr/share/unimaps/`, `/usr/share/consolefonts/` and `/usr/share/keymaps/` folders

I tried this:


```
diff --git a/sdk_container/src/third_party/coreos-overlay/profiles/coreos/targets/generic/prod/make.defaults b/sdk_container/src/third_party/coreos-overlay/profiles/coreos/targets/generic/prod/make.defaults
index 6a34f638ba..2dc5af6747 100644
--- a/sdk_container/src/third_party/coreos-overlay/profiles/coreos/targets/generic/prod/make.defaults
+++ b/sdk_container/src/third_party/coreos-overlay/profiles/coreos/targets/generic/prod/make.defaults
@@ -61,3 +61,13 @@ INSTALL_MASK="${INSTALL_MASK}
   /sbin/ebtables-save
   /sbin/xfs_scrub_all
 "
+
+# remove unnecessary files from the sys-apps/kbd module.
+INSTALL_MASK="${INSTALL_MASK}
+  /usr/share/keymaps
+  -/usr/share/keymaps/include
+  -/usr/share/keymaps/i386/include
+  -/usr/share/keymaps/i386/qwerty/us.map.gz
+  /usr/share/consolefonts
+  /usr/share/unimaps
+"

```

But it is not enough, because it will do changes in the /usr partition, not on the initrd. So the previous way of trying this, with a custom dracut i18n module is required. Will continue by readding the custom module with the above required files.

pothos:
I think it could work if you use `sdk_container/src/third_party/coreos-overlay/profiles/coreos/targets/generic/make.defaults` because then dracut won't be able to find these files in /build/amd64-usr and thus they don't get included in neither the initrd nor the production image.

ader1990:
> I think it could work if you use `sdk_container/src/third_party/coreos-overlay/profiles/coreos/targets/generic/make.defaults` because then dracut won't be able to find these files in /build/amd64-usr and thus they don't get included in neither the initrd nor the production image.

I tried this and the folders were still present in the initrd.

I have tried another way of doing this too, using PKG_INSTALL_MASK:
```
--- a/sdk_container/src/third_party/coreos-overlay/profiles/coreos/base/make.defaults
+++ b/sdk_container/src/third_party/coreos-overlay/profiles/coreos/base/make.defaults
@@ -102,6 +102,14 @@ INSTALL_MASK="

 # Prevent binaries from being installed to rootfs from binary packages
 PKG_INSTALL_MASK="
+  /usr/share/keymaps
+  -/usr/share/keymaps/include
+  -/usr/share/keymaps/i386/include
+  -/usr/share/keymaps/i386/qwerty/us.map.gz
 "
```

The binaries were not present anymore, but what happened, because the "owner" of moving the binaries is the https://github.com/dracutdevs/dracut/blob/master/modules.d/10i18n/module-setup.sh, the 10i18n/module-setup.sh made a symlink of /usr/share/keymaps to /keymaps, probably because this condition here was not valid anymore https://github.com/dracutdevs/dracut/blob/master/modules.d/10i18n/module-setup.sh#L263 which lead to creating the symlinks by the line here: https://github.com/dracutdevs/dracut/blob/master/modules.d/10i18n/module-setup.sh#L134.

At this moment, I think falling back to the custom dracut module looks like a better way.
Any thoughts on this @pothos ?

Thanks.

ader1990:
Ready for review, tested and working with https://github.com/flatcar/scripts/pull/1679.

sayanchowdhury (approved):

sayanchowdhury:
Looks good, but still would let @pothos review.

pothos (approved):

PR #88, opened 2024-02-27
ignition-setup: Keep looking at old place for PXE OEM initrd contents

pothos:
The PXE OEM contents get supplied by a custom initrd the user sets up. This always used /usr/share/oem for the custom contents and we should stick to that instead of looking at /oem.
Revert the breaking change (it was not documented, hence I don't think someone picked this up).


jepio (approved):

krnowak (commented):
Post merge review:

Whoops.

But also: seems like the feature isn't really used then.

PR #89, opened 2024-03-12
initrd-setup-root-after-ignition: Ensure /etc/flatcar/sysext exists

jepio:
Nothing guarantees that this folder is created, so we need to mkdir it before moving the sysext into it. Otherwise the user is forced to create it with ignition when specifying sysexts to install in
/etc/flatcar/enabled-sysext.conf.

### DIFF
--- a/dracut/99setup-root/initrd-setup-root-after-ignition
+++ b/dracut/99setup-root/initrd-setup-root-after-ignition
@@ -134,6 +134,10 @@ for NAME in $(grep -h -o '^[^#]*' /sysroot/etc/flatcar/enabled-sysext.conf /sysr
     echo "Did not find ${ACTIVE_EXT}" >&2
     systemctl start --quiet systemd-networkd systemd-resolved
     download_and_verify "flatcar-${NAME}.raw"
+    # Note: This folder is created opaque since it does not exist in the underlay
+    # For cases where we ship it later, it's best to add a "attr" call in flatcar-postinst
+    # because there are many ways how the folder might have been created
+    mkdir -p "/sysroot/etc/flatcar/sysext/"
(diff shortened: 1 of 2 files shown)


dracut/99setup-root/initrd-setup-root-after-ignition
```diff
@@ -134,6 +134,7 @@ for NAME in $(grep -h -o '^[^#]*' /sysroot/etc/flatcar/enabled-sysext.conf /sysr
     echo "Did not find ${ACTIVE_EXT}" >&2
     systemctl start --quiet systemd-networkd systemd-resolved
     download_and_verify "flatcar-${NAME}.raw"
+    mkdir -p "/sysroot/etc/flatcar/sysext/"
```
pothos, line 140: For this folder and the others such as `/sysroot/etc/flatcar/oem-sysext/` that we don't have in the underlay we could add a comment here that creating it makes it an opaque overlay folder. This means that the underlay won't be used if we want to ship something by default in the future (which we shouldn't because we can also read files from /usr).
```suggestion
    # Note: This folder is created opaque since it does not exist in the underlay
    # For cases where we ship it later, it's best to add a "attr" call in flatcar-postinst
    # because there are many ways how the folder might have been created
    mkdir -p "/sysroot/etc/flatcar/sysext/"
```

pothos (approved):

dracut/99setup-root/initrd-setup-root-after-ignition
jepio: added the comment

jepio:
I've pushed two more changes, can you take another look @pothos ?

pothos (approved):

PR #90, opened 2024-03-16
Add service for root disk unlocking

simoncampion:
# Add service for root disk unlocking

Currently, root disk unlocking in the initramfs has to be triggered by setting a kernel argument `rd.luks...` in the Ignition config. This has two disadvantages. First, it would be nicer for the users if they did not have to set a kernel argument to make root disk unlocking work. Second, during first boot, it leads to an unnecessary systemd-cryptsetup unit that (at least sometimes) fails. While this failing unit causes no harm, it does make the tests fail and might confuse users.

The `systemd-cryptsetup-generator` does not generate because we don't have appropriate entries in `/etc/crypttab` in the initramfs. This PR adds a service that works around this issue by generating a `crypttab` file, calling `systemd-cryptsetup-generator` on that crypttab file to generate a `systemd-cryptsetup` service to decrypt the root, and then starting that service.

## How to use

Root disk decryption set up with `storage.luks` in Ignition should work without passing special kernel arguments in the Ignition config.

## Testing done
 I built the amd64 image based on [this scripts PR](https://github.com/flatcar/scripts/pull/1787) and ran the TPM and Tang tests as modified in [this bootengine PR](https://github.com/flatcar/mantle/pull/511).

```
mantle on remove-luks-cmdline ❯ sudo ./bin/kola run -k cl -p qemu --board amd64-usr --qemu-image ./root-unlock-service.img cl.tpm.* cl.tang.*
2024-03-24T12:54:24Z kola: Using "3916.0.0+nightly-20240321-2100-1-gee375fb003" as version to filter tests...
=== RUN   cl.tpm.root
=== RUN   cl.tpm.nonroot
=== RUN   cl.tang.root
=== RUN   cl.tang.nonroot
--- PASS: cl.tpm.root (61.14s)
--- PASS: cl.tang.nonroot (54.54s)
        tang.go:149: Started tang on 10.0.0.1:44339
--- PASS: cl.tang.root (68.06s)
        tang.go:149: Started tang on 10.0.0.1:43983
--- PASS: cl.tpm.nonroot (54.53s)
PASS, output in _kola_temp/qemu-2024-03-24-1354-761048
```

I failed to get cross-compilation to work and hence have not run the arm64 tests yet. Running CI on the scripts PR will build the arm64 image and I can then run the updated tests against that image.


simoncampion:
It turns out that the systemd-generators in the initramfs run very early, even before device files for block devices are created in `/dev/`. So far, I couldn't come up with a way for the generator to figure out the UUID of the ROOT partition under these circumstances. Any ideas would be very welcome.

If we decide that it's not feasible to implement this generator, we could instead implement a systemd service that runs in the initramfs or even during late user space boot. That service would determines the UUID of the ROOT partition and add `rd.luks.uuid=$uuid` to the kernel cmdline parameters in `grub.cfg`. It would bring the same benefits as the generator: the user would not need to specify the `rd.luks` parameter in their Ignition config, and we would ensure that the cmdline parameter is not yet present during the first boot.

dracut/10root-unlock-generator/root-unlock-generator
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+UNIT_DIR="$1"
+
+# On first boot, Ignition will set up encryption for the ROOT partition, if configured, and leave it unencrypted.
+# Thus, we do not need systemd-cryptsetup unit to decrypt the partition during the first boot.
+if grep -q "flatcar.first_boot" /proc/cmdline ; then
```
pothos: Do we need this or is this precaution? I would think that the generator runs before Ignition formatted the disk, so that it wouldn't detect it anyway, or?

Edit: It's actually good to skip on first boot because we need to make sure that Ignition re-runs would be able to reformat/reencrypt the disk which wouldn't be possible if it's also open at the same time.

pothos:
About the timing; generators may run multiple times. We could also have a look at how the systemd gpt auto generator manages to "wait" for the device.

simoncampion:
> We could also have a look at how the systemd gpt auto generator manages to "wait" for the device.

I might be reading the code wrong, but I think the systemd-gpt-auto-generator does not itself wait for device files to appear. Instead, it generates a cryptsetup service that waits for a device file to appear, which is later created by udev (see [here](https://github.com/systemd/systemd/blob/main/src/gpt-auto-generator/gpt-auto-generator.c#L691)). We could adapt that strategy by adding a systemd service that decrypts the ROOT partition once it's available, but the whole idea was to just piggy-back on the cryptsetup-generator rather than writing our own decryption service.

> About the timing; generators may run multiple times. 

What did you have in mind here? One option that might work (not sure whether that's what you thought of) is to trigger re-running of generators by adding a service that runs `systemctl daemon-reload`. If we do this after udev, the generator I wrote would probably work because all the device files are present. However, I'm not sure running `systemctl daemon-reload` is a good idea, since this is a global operation and might have unintended side-effects.

pothos:
Thanks, thinking about this I guess we can try to define a helper service that runs the logic to invoke the generator. Since we know what service the generator will write out, we can explicitly start this generated service with `systemctl start`. It's best to avoid the daemon-reload and we can do so by calling the generator with `/etc/systemd/system/` as directory to write to.

dracut/10root-unlock-generator/root-unlock-generator
pothos: When we move this shell logic into a helper service, we can let the service be wanted by `ignition-subsequent.target` so that it only runs on second boot. This allows us to drop the duplication of the first-boot cmdline detection.

simoncampion:
> Thanks, thinking about this I guess we can try to define a helper service that runs the logic to invoke the generator. Since we know what service the generator will write out, we can explicitly start this generated service with `systemctl start`. It's best to avoid the daemon-reload and we can do so by calling the generator with `/etc/systemd/system/` as directory to write to.

Sounds good :+1: I implemented this approach. Let me know what you think.

dracut/10root-unlock-generator/root-unlock-generator
simoncampion: Makes sense. In the updated version, the service is `WantedBy=ignition-subsequent.target`.

dracut/31decrypt-root/decrypt-root.service
```diff
@@ -0,0 +1,16 @@
...
+
+DefaultDependencies=no
+# If systemd-networkd.service is used in the initramfs, then we should run after it, since the ROOT partition
+# might use network-bound encryption.
+After=systemd-networkd.service dev-disk-by\x2dpartlabel-ROOT.device
+Wants=dev-disk-by\x2dpartlabel-ROOT.device
+Before=dracut-initqueue.service
+
+[Service]
+Type=oneshot
```
pothos, line 12:
```suggestion
Type=oneshot
RemainAfterExit=yes
```
Not sure if we would hit this but that can prevent the service from being executed more that once

dracut/31decrypt-root/decrypt-root
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
```
pothos, line 3:
```suggestion
set -euo pipefail
```
Would be good to enable error handling. The below line would need to change a bit, though, but in most cases it helps to discover unwanted behavior.

dracut/31decrypt-root/decrypt-root
```diff
@@ -0,0 +1,22 @@
...
+# Check if there is a partition labeled ROOT and, if so, determine its UUID.
+match=$(lsblk -o UUID,PARTLABEL | grep -w ROOT)
```
pothos:
```suggestion
match=$(lsblk --noheadings -o UUID,PARTLABEL | { grep -m1 -w ROOT || true ; })
```
Allow to continue script execution if nothing is found, and also show only one entry.

dracut/31decrypt-root/decrypt-root
```diff
@@ -0,0 +1,22 @@
...
+if [[ $? -ne 0 ]]; then
```
pothos:
```suggestion
if [[ "${match}" = "" ]]; then
```

dracut/31decrypt-root/decrypt-root
```diff
@@ -0,0 +1,22 @@
...
+
+# Check if the partition labeled ROOT is a LUKS device.
+IFS= read -r -n "4" header < /dev/disk/by-uuid/$uuid
+if [[ "$header" != "LUKS" ]]; then
+    exit 0
+fi
+
+# Generate a systemd-cryptsetup unit to decrypt the root partition.
+# We call the LUKS device `rootencrypted`, no matter what it was called in the Ignition configuration.
+echo "rootencrypted UUID=$uuid none luks" > /tmp/crypttab
```
pothos:
```suggestion
echo "rootencrypted UUID=$uuid none luks,tpm2-device=auto" > /tmp/crypttab
```
Would be nice to do this for systemd-cryptsetup to also try it's luck based on the LUKS metadata. I didn't manage yet to get it work but it should be possible.

dracut/31decrypt-root/decrypt-root
```diff
@@ -0,0 +1,22 @@
...
+
+# Check if there is a partition labeled ROOT and, if so, determine its UUID.
+match=$(lsblk -o UUID,PARTLABEL | grep -w ROOT)
+if [[ $? -ne 0 ]]; then
+    exit 0
+fi
+uuid="${match%% *}"
+
+# Check if the partition labeled ROOT is a LUKS device.
+IFS= read -r -n "4" header < /dev/disk/by-uuid/$uuid
```
pothos, line 12: This works for LUKSv1 but somehow not for LUKSv2

pothos: Ah, no, I did it wrong, seems to be working well.

dracut/31decrypt-root/decrypt-root
simoncampion: I added it. (It should not be needed because `clevis` is ready to respond to systemd-cryptsetup's prompts, but it also shouldn't hurt.)

dracut/31decrypt-root/decrypt-root
simoncampion: That's a good point. I added this and made the changes you suggested to the subsequent line to ensure that commands do not fail in normal cases.

simoncampion:
I added the service under `31decrypt-root/` but I'm uncertain whether 31 is the most appropriate number here because I'm uncertain about what the logic behind the names of the various dracut modules in `dracut/` is. Let me know if a different number prefix would be more fitting.

dracut/31decrypt-root/decrypt-root
pothos: Thanks, I'll make a test run. My hope with tpm2-device=auto is for systemd-cryptenroll to also work.

simoncampion: There were failing Tang and TPM tests on arm64 in the test run you triggered, but that's expected because it ran against the mantle main branch. Only the modified tests [here](https://github.com/flatcar/mantle/pull/511) are expected to pass because the `rd.luks.*` cmdline parameters in the current tests create a failing unit in the initramfs.

I downloaded the arm64 image from the pipeline and ran the updated Tang and TPM tests locally. The tests all passed. Could we maybe retrigger the pipeline to run against the tests from the mantle PR?

```
mantle on remove-luks-cmdline ❯ sudo ./bin/kola run --board arm64-usr -v -k -b cl -p qemu --qemu-image ./flatcar_production_qemu_uefi_image.img --qemu-bios=./flatcar_production_qemu_uefi_efi_code.fd cl.tang.* cl.tpm.*
2024-03-25T19:28:50Z cli: Started logging at level INFO
2024-03-25T19:28:50Z util: dnsmasq[24664]: started, version 2.89 cachesize 150
2024-03-25T19:28:50Z util: dnsmasq[24664]: compile time options: IPv6 GNU-getopt DBus no-UBus no-i18n IDN DHCP DHCPv6 no-Lua TFTP conntrack ipset nftset auth cryptohash DNSSEC loop-detect inotify dumpfile
2024-03-25T19:28:50Z util: dnsmasq[24664]: warning: no upstream servers configured
2024-03-25T19:28:50Z util: dnsmasq-dhcp[24664]: DHCP, static leases only on 10.0.0.1, lease time 1h
2024-03-25T19:28:50Z util: dnsmasq-dhcp[24664]: DHCPv4-derived IPv6 names on fd00::
2024-03-25T19:28:50Z util: dnsmasq-dhcp[24664]: router advertisement on fd00::
2024-03-25T19:28:50Z util: dnsmasq-dhcp[24664]: IPv6 router advertisement enabled
2024-03-25T19:28:50Z util: dnsmasq[24664]: cleared cache
2024-03-25T19:28:50Z network/ntp: Started NTP server on [::]:123
2024-03-25T19:28:50Z kola: Creating cluster to check semver...
2024-03-25T19:30:43Z network/ntp: Recieved NTP request from 10.0.0.2:44469
2024-03-25T19:31:15Z network/ntp: Recieved NTP request from 10.0.0.2:60090
2024-03-25T19:31:20Z kola: Using "3916.0.0+nightly-20240321-2100-1-g448d7c8cb3" as version to filter tests...
=== RUN   cl.tpm.nonroot
=== RUN   cl.tang.root
=== RUN   cl.tpm.root
=== RUN   cl.tang.nonroot
2024-03-25T19:34:04Z network/ntp: Recieved NTP request from 10.0.0.3:36340
2024-03-25T19:34:32Z kola/cluster: SSH: running command: lsblk -o NAME,LABEL
2024-03-25T19:34:33Z kola/cluster: SSH: running command: lsblk --json
2024-03-25T19:34:36Z network/ntp: Recieved NTP request from 10.0.0.3:58827
2024-03-25T19:36:34Z network/ntp: Recieved NTP request from 10.0.0.3:39558
2024-03-25T19:36:57Z kola/cluster: SSH: running command: lsblk -o NAME,LABEL
2024-03-25T19:36:59Z kola/cluster: SSH: running command: lsblk --json
--- PASS: cl.tpm.nonroot (342.27s)
2024-03-25T19:39:44Z network/ntp: Recieved NTP request from 10.0.0.4:53216
2024-03-25T19:40:12Z kola/cluster: SSH: running command: lsblk -o NAME,LABEL
2024-03-25T19:40:13Z kola/cluster: SSH: running command: lsblk --json
2024-03-25T19:40:16Z network/ntp: Recieved NTP request from 10.0.0.4:35474
2024-03-25T19:42:20Z network/ntp: Recieved NTP request from 10.0.0.4:38001
2024-03-25T19:42:43Z kola/cluster: SSH: running command: lsblk -o NAME,LABEL
2024-03-25T19:42:45Z kola/cluster: SSH: running command: lsblk --json
--- PASS: cl.tang.nonroot (346.22s)
        tang.go:149: Started tang on 10.0.0.1:45109
2024-03-25T19:45:26Z network/ntp: Recieved NTP request from 10.0.0.5:58449
2024-03-25T19:45:58Z network/ntp: Recieved NTP request from 10.0.0.5:57068
2024-03-25T19:45:59Z kola/cluster: SSH: running command: lsblk -o NAME,LABEL
2024-03-25T19:46:01Z kola/cluster: SSH: running command: lsblk --json
2024-03-25T19:48:21Z kola/cluster: SSH: running command: lsblk -o NAME,LABEL
2024-03-25T19:48:22Z kola/cluster: SSH: running command: lsblk --json
--- PASS: cl.tpm.root (337.51s)
2024-03-25T19:53:43Z network/ntp: Recieved NTP request from 10.0.0.6:56679
2024-03-25T19:54:11Z kola/cluster: SSH: running command: lsblk -o NAME,LABEL
2024-03-25T19:54:13Z kola/cluster: SSH: running command: lsblk --json
2024-03-25T19:54:15Z network/ntp: Recieved NTP request from 10.0.0.6:56049
2024-03-25T19:56:21Z network/ntp: Recieved NTP request from 10.0.0.6:38254
2024-03-25T19:56:51Z kola/cluster: SSH: running command: lsblk -o NAME,LABEL
2024-03-25T19:56:53Z kola/cluster: SSH: running command: lsblk --json
2024-03-25T19:56:53Z network/ntp: Recieved NTP request from 10.0.0.6:41550
--- PASS: cl.tang.root (510.97s)
        tang.go:149: Started tang on 10.0.0.1:34615
PASS, output in _kola_temp/qemu-2024-03-25-2028-24642
```

ader1990:
> I added the service under `31decrypt-root/` but I'm uncertain whether 31 is the most appropriate number here because I'm uncertain about what the logic behind the names of the various dracut modules in `dracut/` is. Let me know if a different number prefix would be more fitting.

Hello @simoncampion, the only usable systemd command for this scenario is `systemd-analyze plot`. I used it succesfully during my kmod static nodes PR, to be able to see properly the systemd execution tree during initrd and after switch root.

Should look like this and is very helpful to see the order 
plot-base

pothos (approved):

dracut/31decrypt-root/decrypt-root
pothos: I see you ran the arm64 tests, so I guess it's good then without doing one more run here.

simoncampion:
@ader1990 Thanks for the hint! That's a neat command.

PR #91, opened 2024-03-22
Draft for Proxmox support

pothos:
This makes use of https://github.com/coreos/afterburn/pull/1023 to set up any static networking from the initrd (for Ignition) and the hostname (early enough so that Ignition could overwrite it).

## Testing done

Tested with Flatcar community: https://github.com/flatcar/Flatcar/discussions/1573#discussioncomment-11170022

### DIFF
--- a/dracut/30ignition/flatcar-afterburn-network.service
+++ b/dracut/30ignition/flatcar-afterburn-network.service
@@ -1,5 +1,5 @@
 [Unit]
-Description=Flatcar DigitalOcean Network Agent
+Description=Flatcar Afterburn network service
 DefaultDependencies=false
 Before=initrd.target
 After=systemd-networkd.service initrd-root-fs.target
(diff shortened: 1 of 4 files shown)


pothos:
I think we have to close this because this relies on cloud-init userdata and because Flatcar will run coreos-cloudinit anyway if it doesn't see Ignition user-data we would have it processed twice. For the network config that can be quite confusing because the afterburn network unit would only support a subset.

pothos:
> I think we have to close this because this relies on cloud-init userdata and because Flatcar will run coreos-cloudinit anyway if it doesn't see Ignition user-data we would have it processed twice. For the network config that can be quite confusing because the afterburn network unit would only support a subset.

Or not because the configdrive has two files, and afterburn could be used for `network-config` while we leave `user-data` to Ignition/coreos-cloudinit (through the configdrive unit trigger).

PR #92, opened 2024-03-26
ignition: fixes hyperv oem support

ader1990:
Upstream Hyper-V  ignition support requires `hv_utils` module.

See: https://github.com/coreos/ignition/commit/d304850c3d3696822bc05e0833ee4b27df9d7a38

Ignition initrd is failing here: https://github.com/coreos/ignition/blob/v2.18.0/internal/providers/hyperv/kvp.go#L50.
Flatcar kernel is built with hv_utils kernel module, but the module is not present in initrd  /lib/modules because it has not been installed.

### DIFF
--- a/dracut/30ignition/module-setup.sh
+++ b/dracut/30ignition/module-setup.sh
@@ -184,3 +184,9 @@ EOF
     done
 
 }
+
+# See: https://github.com/coreos/ignition/commit/d304850c3d3696822bc05e0833ee4b27df9d7a38
+installkernel() {
+     # required by hyperv platform to read kvp from the kernel
+     instmods -c hv_utils
+}


jepio (approved):

ader1990:
Found out that PR https://github.com/flatcar/scripts/pull/1680 introduced https://github.com/coreos/ignition/commit/d304850c3d3696822bc05e0833ee4b27df9d7a38, which means the Hyper-V VHD is broken due to not adding this change to the bootengine.

ader1990:
Boot partition used space: `62620 KB`, which is a normal increase, as the hv_utils.ko.xz size is 20K.

ader1990:
Tested locally with this patch and Hyper-V vhdx is booting correctly now (ignition is running correctly during initrd stage now).

PR #94, opened 2024-04-11
afterburn: set hostname for hetzner OEM

apricote:
# Use Hostname from metadata service for Hetzner OEM

Enables the existing afterburn service `flatcar-metadata-hostname.service` for the Hetzner OEM. This will read the hostname of the server from the Hetzner Cloud metadata service in the init process, and configure the server to use this hostname.

## How to use

Changes are pulled into `flatcar/scripts` in https://github.com/flatcar/scripts/pull/1880. This also explains how to use/test them.

## Testing done

I have referenced this commit in `flatcar/scripts` overlay and built a test image. A server created from this image used the correct hostname.

### DIFF
--- a/dracut/30ignition/flatcar-metadata-hostname.service
+++ b/dracut/30ignition/flatcar-metadata-hostname.service
@@ -33,6 +33,7 @@ ConditionKernelCommandLine=|flatcar.oem.id=vultr
 # Addition:
 ConditionKernelCommandLine=|coreos.oem.id=packet
 ConditionKernelCommandLine=|flatcar.oem.id=packet
+ConditionKernelCommandLine=|flatcar.oem.id=hetzner
 
 OnFailure=emergency.target
 OnFailureJobMode=isolate


dracut/30ignition/flatcar-metadata-hostname.service
```diff
@@ -33,6 +33,8 @@ ConditionKernelCommandLine=|flatcar.oem.id=vultr
 # Addition:
 ConditionKernelCommandLine=|coreos.oem.id=packet
 ConditionKernelCommandLine=|flatcar.oem.id=packet
+ConditionKernelCommandLine=|coreos.oem.id=hetzner
```
tormath1:
```suggestion
```

Same reason as in https://github.com/flatcar/scripts/pull/1880#discussion_r1561377054

apricote: Removed in b8417f39694dc02ff1bdf814e1dd00b0f7aba3d3

PR #97, opened 2024-05-31
dracut: Insert dm_verity kernel module for veritysetup

krnowak:
For some reason it stopped being inserted.

Tested as a part of https://github.com/flatcar/scripts/pull/1982


krnowak:
Replaced by https://github.com/flatcar/init/pull/122.

PR #98, opened 2024-08-21
decrypt-root: Rework around a udev rule to fix slow PXE/ISO boots

### LINKED ISSUE #1514: Boot time for ISO image `3975.2.0` release has increased to ~1m30s
## Description

The latest stable release, `3975.2.0`, is taking a considerably longer time to boot compared to previous releases. 

E.g.

```
core@localhost ~ $ systemd-analyze
Startup finished in 1.787s (kernel) + 1min 31.064s (initrd) + 4.103s (userspace) = 1min 36.956s
multi-user.target reached after 4.103s in userspace.
```

vs. the previous release:

```
Startup finished in 1.309s (kernel) + 2.089s (initrd) + 4.278s (userspace) = 7.677s
multi-user.target reached after 4.254s in userspace.
```

This has resulted in a delay at boot of roughly 1min 30sec compared to the previous release.

## Impact

When building images with [image-builder](https://github.com/kubernetes-sigs/image-builder/) the default boot wait time is 60s after which Packer will start typing out the boot command. As this happens before the OS has finished booting the `flatcar-install` script isn't actually run correctly and can result in the OS not being correctly set up in an environment such as VSphere (it doesn't install the vmtools for example). With image-builder this then results in it hanging at `Waiting for IP...` as vcenter is unable to retrieve the IP from the guest OS as no vmtools have been installed.

## Environment and steps to reproduce

1. **Set-up**: Flatcar `3975.2.0` on VSphere via `image-builder`
2. **Task**: Packer stuck trying to get the IP address of the VM
3. **Action(s)**: 
  a. With image-builder run: `make build-node-ova-vsphere-flatcar`
  b. The `3975.2.0` image is used to launch a new VM (`https://stable.release.flatcar-linux.net/amd64-usr/3975.2.0/flatcar_production_iso_image.iso`)
  c. The VM boots but vcenter fails to retrieve the IP address due to lack of vmtools on the guest OS
4. **Error**: `Waiting for IP...`

## Expected behavior

The OS boots in under 60 seconds. Ideally back down below 10s like the previous release.

## Additional information

This issue was previously about `open-vm-tools` not being installed but after debugging with @tormath1 on Slack ([see thread](https://kubernetes.slack.com/archives/C03GQ8B5XNJ/p1723196516253269)) we learnt that it was actually due to the delay in boot that was causing the problem so I updated the issue to reflect that.

chewi:
# decrypt-root: Rework around a udev rule to fix slow boots

PXE and ISO boots were taking over 1m30s because decrypt-root.service was waiting for the ROOT partition to appear just in case it needed to be decrypted.

Making this service conditional on non-PXE/ISO boots was considered, but users may well want an encrypted ROOT when booting using these methods. Adding conditions didn't avoid the delay caused by the `Wants` directive in any case.

This change makes decrypt-root.service more reactive by having it triggered by a udev rule matching a LUKS-encrypted partition named ROOT. systemd knows to wait for `/dev/mapper/rootencrypted` to appear before attempting to switch root.

Adding `After=ignition-disks.service` ensures that the service does not try to decrypt ROOT before Ignition wipes it again due to `flatcar.first_boot=1` being given.

The service itself has been simplified. crypttab is just as happy with a block device path, so there is no need to determine the UUID. The udev rule only matches LUKS-encrypted partitions, so there is no need to check the header either. The rest is simple enough to include in the service unit directly, removing the need for a separate shell script. systemd-cryptsetup-generator can read the crypttab from stdin, avoiding the creation of a temporary file. It is also important to call daemon-reload before attempting to start the generated unit.

Closes: https://github.com/flatcar/Flatcar/issues/1514

## How to use

Images are present on [bincache](https://bincache.flatcar-linux.net/images/amd64/9999.9.9+decrypt-root-udev-2/). A regular ISO or PXE boot with QEMU should only take a few seconds. You can experiment with encryption using these:

```yaml
variant: flatcar
version: 1.0.0
storage:
  disks:
  - device: /dev/vda
    wipe_table: true
    partitions:
    - label: ROOT
  luks:
    - name: rootencrypted
      wipe_volume: true
      device: "/dev/disk/by-partlabel/ROOT"
  filesystems:
    - device: /dev/mapper/rootencrypted
      format: ext4
      label: ROOT
systemd:
  units:
    - name: cryptenroll-helper.service
      enabled: true
      contents: |
        [Unit]
        ConditionFirstBoot=true
        OnFailure=emergency.target
        OnFailureJobMode=isolate
        [Service]
        Type=oneshot
        RemainAfterExit=yes
        ExecStart=systemd-cryptenroll --tpm2-device=auto --unlock-key-file=/etc/luks/rootencrypted --wipe-slot=0 --tpm2-pcrs= /dev/disk/by-partlabel/ROOT
        ExecStart=rm /etc/luks/rootencrypted
        [Install]
        WantedBy=multi-user.target
```

```
qemu-img create -f qcow2 disk.img 6g
./flatcar_production_pxe.sh -i ignition.json -I disk.img -T swtpm -- -append 'flatcar.autologin flatcar.first_boot=1 root=/dev/mapper/rootencrypted'
```

You can use the regular QCOW2 image by removing the `disks` section so that it encrypts the existing ROOT partition.

## Testing done

A lot of manual testing around the above. The Jenkins packages_all_arches job passed fine.

- [ ] Changelog entries added in the respective `changelog/` directory (user-facing change, bug fix, security fix, update) -- **I'll add this to flatcar/scripts**
- [X] Inspected CI output for image differences: `/boot` and `/usr` size, packages, list files for any missing binaries, kernel modules, config files, kernel modules, etc.


dracut/31decrypt-root/decrypt-root.service
```diff
@@ -4,14 +4,14 @@ Description=Generate and execute a systemd-cryptsetup service to decrypt the ROO
 DefaultDependencies=no
 # If systemd-networkd.service is used in the initramfs, then we should run after it, since the ROOT partition
 # might use network-bound encryption.
-After=systemd-networkd.service dev-disk-by\x2dpartlabel-ROOT.device
+After=systemd-networkd.service dev-disk-by\x2dpartlabel-ROOT.device ignition-disks.service
 Wants=dev-disk-by\x2dpartlabel-ROOT.device
-Before=dracut-initqueue.service
```
krnowak, line 9: Could you maybe update the commit message to tell why `Before=dracut-initqueue.service` and `[Install] WantedBy=ignition-subsequent.target` lines were dropped?

krnowak (approved):
Asking only for a commit message update. Otherwise looks good. I assume that Jenkins job you mentioned was http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/4534/cldsv/, right?

PR #103, opened 2025-02-27
ignition: Fix PXE by using mount namespace rather than remounting RO

chewi:
Remounting read-only was failing due to systemd v256 holding a read-write file descriptor on systemd-executor. We don't need to do this inside a namespace. We don't need to unmount the OEM partition either.

[Jenkins](http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/5478/cldsv/) is all green. I included Equinix Metal on amd64 and arm64.

### DIFF
--- a/dracut/30ignition/ignition-generator
+++ b/dracut/30ignition/ignition-generator
@@ -137,6 +137,7 @@ OnFailureJobMode=isolate
 [Service]
 Type=oneshot
 RemainAfterExit=yes
+MountFlags=slave
 ExecStart=/usr/sbin/ignition-setup ${nopxe:+normal} ${pxe:+pxe}
 EOF
 
--- a/dracut/30ignition/ignition-setup.sh
+++ b/dracut/30ignition/ignition-setup.sh
@@ -5,16 +5,14 @@ set -e
 
 # remount read write for Ignition required steps
 mount -o remount,rw /usr
-# remount read only after Ignition required steps finalized
-trap 'mount -o remount,ro /usr' EXIT
+# we run with MountFlags=slave, so the rw remount is not propagated outside the
+# unit - no need to remount ro or umount the OEM partition
 
 case "$1" in
 normal)
     src=/mnt/oem
     mkdir -p "${src}"
     mount /dev/disk/by-label/OEM "${src}"
-    # retry-umount may not be necessary, but be cautious
-    trap 'retry-umount "${src}"' EXIT
     # Workaround, "chmod" is not available
     cp -a /bin/cat /bin/is-live-image
     printf '#!/bin/sh\nexit 1\n' > /bin/is-live-image


ader1990:
Hello, can you share the bug link or description that this PR fixes?

chewi:
It was only discussed on [Matrix](https://matrix.to/#/!SakMJmpDOgVMlGukUp:matrix.org/$HZyYi22Py7NR-sLfv0grCiVTDNgezP_ejx2ecE0QO7o?via=matrix.org&via=gaf.fs.lmu.de&via=hcrypt.net). The above basically sums it up though. This went unnoticed initially because we don't run tests on EM by default, and it's known to be flaky sometimes.

ader1990:
> It was only discussed on [Matrix](https://matrix.to/#/!SakMJmpDOgVMlGukUp:matrix.org/$HZyYi22Py7NR-sLfv0grCiVTDNgezP_ejx2ecE0QO7o?via=matrix.org&via=gaf.fs.lmu.de&via=hcrypt.net). The above basically sums it up though. This went unnoticed initially because we don't run tests on EM by default, and it's known to be flaky sometimes.

Can you please provide the context and the actual error logs / files on this matter here? A PR needs to have a link to an issue or a clear description of what the issue issue / error logs.

chewi:
Both amd64 and arm64 CI tests totally failed. The logs were full of errors like this:

```
04:50:13  2025-02-25T04:50:12Z platform/api/equinixmetal: 7987be91-f02c-4a6b-a580-04b280bcbdd5 console session failed: Process exited with status 1
05:38:34  2025-02-25T05:38:27Z platform/machine/equinixmetal: Retrying to provision a machine after error: "timed out waiting for flatcar-install: dial tcp 147.75.203.3:9: i/o timeout"
05:38:42  2025-02-25T05:38:41Z platform/api/equinixmetal: abddd3ff-d46d-4c2b-b314-7b9db5a5c66f console session failed: Process exited with status 1
06:27:02  2025-02-25T06:26:56Z platform/machine/equinixmetal: Retrying to provision a machine after error: "timed out waiting for flatcar-install: dial tcp 147.75.90.1:9: i/o timeout"
```

Logging in via the serial console showed that it was stuck in the emergency shell having failed to boot. This was due to ignition-setup.service failing, which showed this:

```
:/root# systemctl --no-pager -o cat status ignition-setup
× ignition-setup.service - Ignition (setup)
     Loaded: loaded (/run/systemd/generator/ignition-setup.service; generated)
     Active: failed (Result: exit-code) since Tue 2025-02-25 16:10:40 UTC; 5min ago
 Invocation: f24e87129324420aa82ca1c79c718092
    Process: 481 ExecStart=/usr/sbin/ignition-setup pxe (code=exited, status=32)
   Main PID: 481 (code=exited, status=32)

Starting ignition-setup.service - Ignition (setup)...
ignition-setup.service: Main process exited, code=exited, status=32/n/a
mount: /usr: mount point is busy.
       dmesg(1) may have more information after failed mount system call.
ignition-setup.service: Failed with result 'exit-code'.
Failed to start ignition-setup.service - Ignition (setup).
ignition-setup.service: Triggering OnFailure= dependencies.
```

Running lsof via the mounted /sysroot/usr showed a read-write file descriptor within /usr, indicated by the `u`.

```
COMMAND   PID USER  FD   TYPE DEVICE SIZE/OFF NODE NAME
systemd     1 root   9u   REG    0,2   137520 1232 /usr/lib/systemd/systemd-executor
```

krnowak (approved):
That's a neat fix, especially that the unit does not do any mounting that should remain after the process is finished.

Aside that, I'm wondering how this thing worked exactly for the "normal" case, where the later `trap 'retry-umount "${src}"' EXIT` clobbers the earlier `trap 'mount -o remount,ro /usr' EXIT`. And also, without the clobbering, the same issue probably would arise on other platforms.

Also, it looks like that the `/usr` was remounted as read-only, but at a later stage? Because I don't remember seeing problems with writeable /usr.

chewi:
Very good points! I had wondered why it only affected PXE, and that totally explains it.

PR #104, opened 2025-04-08
Add badges to README

John15321:
This PR prepends a set of Flatcar-related badges to the README.


chewi (changes requested):
This one has broken the header due to a missing blank line in-between.

John15321:
@chewi Thanks for spotting this :smile: Should be good now

chewi (approved):

PR #105, opened 2025-04-22
Miscellaneous changes needed to use newer Dracut with its sysroot feature

chewi:
# Misc changes needed to use new Dracut with sysroot feature

This must be reviewed in tandem with flatcar/scripts#2837.

Flatcar has lagged behind on Dracut 053 for quite a while. The latest version is currently 106, although it did jump from 060 to 100 when it changed hands.

It's hard to summarise the changes here, so please see the commit messages. The main change is that we now run Dracut with its `--sysroot` argument, removing the need to chroot and jump through hoops to avoid the Portage sandbox in the coreos-kernel ebuild.

Before you ask, the extglob stuff _is_ needed twice, once when the script is parsed, and once when the `install()` function is executed. Bash is weird.

## How to use

This is only usable with my chewi/dracut branch in flatcar/scripts. You will need the latest SDK due to changes in my earlier PR. You should be able to update dracut in the SDK, plus dracut and bootengine in the board root, before building coreos-kernel.

## Testing done

I have done a lot of manual testing and [CI testing](http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/5825/cldsv/), as well as a close inspection of the initrd changes, in tandem with my chewi/dracut branch in flatcar/scripts.

- [X] Inspected CI output for image differences: `/boot` and `/usr` size, packages, list files for any missing binaries, kernel modules, config files, kernel modules, etc.

### DIFF
--- a/dracut/30ignition/module-setup.sh
+++ b/dracut/30ignition/module-setup.sh
@@ -58,6 +58,8 @@ install() {
         clevis-luks-common-functions \
         clevis-luks-unlock \
         pwmake \
+        sort \
+        tail \
         tpm2_create
 
     # Required by s390x's z/VM installation.
@@ -93,6 +95,7 @@ install() {
     inst_simple "$moddir/ignition-generator" \
         "$systemdutildir/system-generators/ignition-generator"
 
+    local x
     for x in "complete" "subsequent" "diskful" "diskful-subsequent"; do
         inst_simple "$moddir/ignition-$x.target" \
             "$systemdsystemunitdir/ignition-$x.target"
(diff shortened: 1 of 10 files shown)


krnowak:
> Before you ask, the extglob stuff is needed twice, once when the script is parsed, and once when the install() function is executed. Bash is weird.

Yeah, the situation is stupid. `extglob` is enabled by default for interactive shells, but not for non-interactive ones. It's annoying.

dracut/30ignition/module-setup.sh
```diff
@@ -129,58 +131,35 @@ install() {
...
-        /usr/local/libexec/clevis-luks-askpass \
-        /usr/libexec/clevis-luks-generic-unlocker \
-        /usr/sbin/setfiles \
-    ; do
-        directory="$(dirname "$executable")"
-        filename="$(basename "$executable")"
-
-        wrapper_name="${filename}-wrapper"
-        cat <<EOF > /tmp/${filename}-wrapper
+    for executable in "${dracutsysrootdir}"{\
```
krnowak, line 139: You probably could add `local executable` a line above. And possibly `local x` for the `for x in "complete"…` loop above too, while at it.

dracut/99flatcar-debloat/module-setup.sh
```diff
@@ -0,0 +1,21 @@
...
+install() {
+    # shellcheck disable=SC2064
+    trap "$(shopt -p extglob)" RETURN
+    shopt -q -s extglob
+
+    # Remove the NSS modules we don't need.
+    rm "${initdir}"/usr/lib*/libnss_!(dns|files|myhostname|resolve|systemd).so*
+
+    # We maybe should include this, but more work is needed for compliance.
+    rm "${initdir}"/usr/lib*/ossl-modules/fips.so
```
krnowak, line 20: This is probably something @tormath1 would be interested to see (if I remember correctly, you were working on fips mode in Flatcar, right?).

chewi: I already spoke to him about it. :slightly_smiling_face: Also, it wasn't being included before, only newer Dracut versions do this.

dracut/30ignition/module-setup.sh
chewi: Good shout, done. Also fixed a couple of other shellcheck issues.

krnowak (approved):

PR #108, opened 2025-08-12
dracut: Reduce size with wrappers

pothos:
The xfs repair binaries are large and the user might need them manually
for the rootfs but at that point we can load them from /usr, too,
through a wrapper script. They are not called automatically (fsck.xfs
does not exist because the journal is enough for regular recovery).

## Testing done

Ran the wrappers and checked that fsck is not done automatically for xfs.

### DIFF
--- a/dracut/30ignition/module-setup.sh
+++ b/dracut/30ignition/module-setup.sh
@@ -144,6 +144,8 @@ install() {
 /usr/bin/jose,\
 /usr/bin/luksmeta,\
 /usr/bin/tpm2,\
+/usr/bin/xfs_db,\
+/usr/bin/xfs_repair,\
 /usr/lib/systemd/systemd-reply-password,\
 /usr/libexec/clevis*\
 }; do


pothos:
The tpm2 binary is not created as wrapper because it seems something copies the real binary again after the dracut Ignition module ran? We need to check this in a follow up.

chewi (approved):
Nice! As you know, I have a plan, but this is a solid win in the meantime.

For tpm2, maybe this whole wrapper setup could move to the debloat module.

pothos:
> For tpm2, maybe this whole wrapper setup could move to the debloat module.

I tried that in this PR first but something still ended up placing the binary on top.

ader1990:
> > For tpm2, maybe this whole wrapper setup could move to the debloat module.
> 
> I tried that in this PR first but something still ended up placing the binary on top.

probably dracut code?

chewi:
It might be Clevis, but I haven't checked.

PR #109, opened 2025-09-02
dracut: move the module ordering to the 50-59 range

jozzsi:
# Follow the recommended dracut module ordering for out-of-tree modules.

For dracut the recommended ordering for out out of tree modules is 50-59 range. The following is a section from dracut documentation:

> Not using the 50-59 range for out of tree dracut modules will likely
> lead to unintended errors in the initramfs generation process as your
> dracut module will either run too early or too late in the generation process.
> You have been warned.


jozzsi:
@chewi - could you please assist how best to approach this work ?

I suspect there needs to be a discussion for the modules with 99 ordering as well, but could be done as a follow-up PR.

My motivation here is try to avoid changing the existing ordering - this is how I landed on the numbers on this PR

chewi:
I was really confused until I noticed which repo this was. :sweat_smile: 

I'm a little nervous about changing this, as Flatcar does more in early boot than most distros. We could give it a spin in CI to see how it goes as we do have a thorough test suite.

However, we have been battling with a lack of disk space in /boot and are currently investigating a total replacement for bootengine that might not involve Dracut. I've had to divert off that task recently, but I hope to come back to it soon. It therefore might not be worth changing this now.

ader1990 (changes requested):
This change is not needed (do not fix what is not broken). In the case of Flatcar, if we want to change the ordering, we need first to come up with the current tree of modules as a visual representation, and see if this change changes that tree of modules.

Note: this change seems automated and part of PRs on multilple projects, see: https://github.com/jozzsi?tab=repositories where there are multiple two branch forks, with the extra branch called `1`, from where the PRs on the upstream projects are done.

ader1990:
@jozzsi please provide the DCO, as it is now required for Flatcar repositories contributions. See https://github.com/flatcar/bootengine/pull/109/checks?check_run_id=49370455074. Thanks!

chewi:
Some extra context is needed here. I have worked with @jozzsi on several Dracut changes, including his recent renumbering of the core modules. I think this was released in v108, which we haven't taken yet. He is rightfully concerned that this might break things here.

jozzsi:
Thanks @chewi for helping out.

> This change is not needed (do not fix what is not broken). 

I understand @ader1990 . I let the Flatcar team handle, so let me close my version of the PR.

ader1990:
> Thanks @chewi for helping out.
> 
> > This change is not needed (do not fix what is not broken).
> 
> I understand @ader1990 . I let the Flatcar team handle, so let me close my version of the PR.

Hello, related to `do not fix what is not broken` - this change **_MAY_** be required in the future, as @chewi said, so it would be great to have it prepared and ready if we switch to the dracut version that might pose problems.

What I was saying is that we need a visual representation or at least a text representation of what is currently in the Dracut Flatcar module order and then another representation with this patch, so that we make sure the modules **_ARE_** in the same order as before.

Feel free to reopen the PR at any time.

Thanks.

ader1990:
> > Thanks @chewi for helping out.
> > > This change is not needed (do not fix what is not broken).
> > 
> > 
> > I understand @ader1990 . I let the Flatcar team handle, so let me close my version of the PR.
> 
> Hello, related to `do not fix what is not broken` - this change **_MAY_** be required in the future, as @chewi said, so it would be great to have it prepared and ready if we switch to the dracut version that might pose problems.
> 
> What I was saying is that we need a visual representation or at least a text representation of what is currently in the Dracut Flatcar module order and then another representation with this patch, so that we make sure the modules **_ARE_** in the same order as before.
> 
> Feel free to reopen the PR at any time.
> 
> Thanks.

Also, the DCO is now a requirement from the CNCF, so the commit needs to be signed to be mergeable (see https://github.com/flatcar/bootengine/pull/109/checks?check_run_id=49370455074).

chewi:
I found this change was needed for Dracut 109, and it worked perfectly following a rebase. Many thanks to @jozzsi. I wasn't able to use this PR because the fork has been deleted, so there's a new one at #116.

PR #110, opened 2025-09-12
Use a minimal initrd to switch to the full initrd stored in /usr

pothos:
The growth of binaries over time and the inclusion of new features
    filled the available boot partition space, so that the kernel+initrd
    almost couldn't fit twice anymore as required for updates. We employed
    workarounds such as wrapper scripts for ignition, afterburn and other
    binaries so that they are loaded from /usr. However, this was still not
    enough and we would have to do the same for (network) kernel modules and
    firmware. To avoid making this ever more complex we can use a dedicated
    initrd focused on loading the full initrd from /usr and then this full
    initrd can use dracut as before and even drop all the workarounds we
    accumulated.

Introduce a busybox init script that prepares a minimal environment,
    has debug toggles and an emergency shell, and only loads the real initrd
    from /usr to switch over to it. Because mdev is not a proper udev
    replacement, some additional scripting is needed. Busybox's modprobe
    can't work with dependencies well and we need the real kmod for that
    (which is also good to guarantee have the same modprobe options set).
    Also, some other busybox commands are often lacking things such as
    loading a kernel module automatically and this has to be done
    explicitly. We still set up dm-verity for /usr so that we have the same
    security properties (The code comes from the bootengine systemd
    generators we have and also covers the PXE boot with a squashfs /usr
    passed from an additional cpio). The real initrd then reuses the mount
    point for /usr, and loads any kernel modules and firmware that wasn't
    loaded already.
    We also have to make the dependencies for parse-ip-for-networkd.service
    a bit more explicit because the removal of the /sysusr mount in the full
    initrd exposed a race condition.

## How to use

With https://github.com/flatcar/scripts/pull/3241

## Testing done

See above


minimal-init
```diff
@@ -0,0 +1,189 @@
...
+  for arg in $(cat /proc/cmdline); do
+    if [[ "${arg%%=*}" == "${name}" ]]; then
+      value="${arg#*=}"
+    fi
+  done
+  echo "${value}"
+}
+
+emergency() {
+  echo "ERROR: The early initrd has failed. To active debug shell breakpoints, boot with rd.earlyshell in the kernel cmdline, and to active tracing, boot with rd.earlytrace" >&2
```
krnowak: Did you mean to use `activate` here?

```suggestion
  echo "ERROR: The early initrd has failed. To activate debug shell breakpoints, boot with rd.earlyshell in the kernel cmdline, and to activate tracing, boot with rd.earlytrace" >&2
```

minimal-init
```diff
@@ -0,0 +1,189 @@
...
+  esac
+  starttime=$(date +%s)
+  while [ "${drive}" = "" ]; do
+    now=$(date +%s)
+    # Timeout of 5 minutes for finding the device
+    # NOTE: Only mdev -d runs as this point and the kernel also can spawn modprobe to load modules.
+    # If problems arise, first make sure that required modules and their deps are actually in the initrd,
+    # but if that's not enough we might even have to trigger the find /sys ... xargs coldplugging
+    # here again every now and then? (Last resort would be to run proper udev, possibly without systemd.)
+    if [ $((now - starttime)) -gt 300 ]; then
```
krnowak: Would it make sense to be able to override the timeout through the cmdline parameters? Dunno, maybe there are platforms, where you need to wait for the disk to show up.

minimal-init
```diff
@@ -0,0 +1,189 @@
...
+# Ported code from the generators
+verityusr=$(cmdline_arg verity.usr)
+usrhash=$(cmdline_arg verity.usrhash)
+
+verityusr=$(find_drive "${verityusr}")
+
+# Only proceed if the source is a path and we have sufficient parameters.
+if echo "${verityusr}" | grep -q "^/" && [ "${usrhash}" != "" ]; then
+  # Hardcoded expected value from the image GPT layout
+  veritysetup --panic-on-corruption --hash-offset=1065345024 open "${verityusr}" usr "${verityusr}" "${usrhash}"
```
krnowak, line 127: Weren't offsets different on amd64 and arm64? Or did I just confuse things?

minimal-init
```diff
@@ -0,0 +1,189 @@
...
+  echo "No mountable /usr partition given (usr='${usr}')" >&2
+  false # Throw error
+fi
+
+debug_sh 3/4: before /sysusr mount
+
+echo "Mounting /usr from ${usr}" >&2
+# mount -t auto only works if btrfs is already loaded
+modprobe btrfs
+mount -t "${usrfstype}" -o "${usrflags}" "${usr}" /sysusr/usr
```
krnowak, line 168: Hm, will `mount` create the `/sysusr/usr` directories for us or it is created elsewhere?

minimal-init
pothos: Yes, typo, thanks

minimal-init
pothos: Yes, we could add a parameter, I thought 5 minutes are very long already but who knows, and it could be handy when testing to set it to a smaller timout

minimal-init
pothos: For the place where the hash is located in the kernel binary, yes, but the verity data offset on disk is the same.

minimal-init
pothos: It's set up in the initrd build step, same for the `realinit` folder. We could also add `mkdir -p` here again?

krnowak: Doing `mkdir -p /sysusr/usr` certainly wouldn't hurt.

krnowak (approved):
Looks good, but I think Chewi will want to have a look too.

minimal-init
pothos: Sure, added it, also for `/realinit`

minimal-init
pothos: Added `rd.earlytimeout`

minimal-init
```diff
@@ -0,0 +1,194 @@
+#!/bin/sh
+set -u
+busybox mount -n -t proc proc /proc
+busybox mount -n -t devtmpfs devtmpfs /dev
+busybox mount -n -t sysfs sysfs /sys
+busybox --install -s
```
chewi, line 6: Perhaps we still need this for subprocesses, but the busybox shell does not need symlinks to invoke built-in tools without prepending them with `busybox`. It will prefer a tool from the PATH if one exists, otherwise it will use the built-in version.

minimal-init
```diff
@@ -0,0 +1,194 @@
...
+debug_sh 1/4: before mdev
+if [ "$(cmdline_arg rd.earlytrace)" != "" ]; then
+  set -x
+fi
+
+mdev -d
+mdev -s
+# Coldplugging but with using /sbin/modprobe (which is kmod) instead of busybox's modprobe
+# because busybox doesn't properly support the globs in modules.alias
+find /sys/ -name modalias -print0 | xargs -0 sort -u | tr '\n' '\0' | xargs -0 /sbin/modprobe -abq || true
```
chewi, line 51: What gets passed to modprobe exactly? When I try this on my systems, I get entries like this, which `modprobe` doesn't understand.

```
of:NvopT(null)Crockchip,rk3566-vop
of:NwatchdogT(null)Crockchip,rk3568-wdtCsnps,dw-wdt
pci:v00001D87d00003566sv00000000sd00000000bc06sc04i00
pci:v00001E0Fd00000009sv00001E0Fsd00000001bc01sc08i02
platform:alarmtimer
```

minimal-init
```diff
@@ -0,0 +1,194 @@
...
+    # here again every now and then? (Last resort would be to run proper udev, possibly without systemd.)
+    if [ $((now - starttime)) -gt "${timeoutsecs}" ]; then
+      echo "ERROR: Timeout waiting for drive: ${ueventline}${blkidmatch}" >&2
+      return 1 # Throw error
+    fi
+    # No "sleep 0.1", so this is rather busy polling
+    if [ "${ueventline}" != "" ]; then
+      drive="$({ grep -s -l -m 1 -r "${ueventline}" /sys/class/block/*/uevent || true; } | cut -d / -f 5)"
+    else
+      drive="$(blkid | { grep -m 1 "${blkidmatch}" || true ; } | cut -d : -f 1 | cut -d / -f 3-)"
```
chewi, line 107: I don't think you need the `|| true` because there is no `set -o pipefail`.

```suggestion
      drive="$(grep -s -l -m 1 -r "${ueventline}" /sys/class/block/*/uevent | cut -d / -f 5)"
    else
      drive="$(blkid | grep -m 1 "${blkidmatch}" | cut -d : -f 1 | cut -d / -f 3-)"
```

minimal-init
```diff
@@ -0,0 +1,194 @@
...
+# Ported code from the generators
+verityusr=$(cmdline_arg verity.usr)
+usrhash=$(cmdline_arg verity.usrhash)
+
+verityusr=$(find_drive "${verityusr}")
+
+# Only proceed if the source is a path and we have sufficient parameters.
+if echo "${verityusr}" | grep -q "^/" && [ "${usrhash}" != "" ]; then
+  # Hardcoded expected value from the image GPT layout
+  veritysetup --panic-on-corruption --hash-offset=1065345024 open "${verityusr}" usr "${verityusr}" "${usrhash}"
```
chewi, line 127: Bottlerocket was able to do this without `veritysetup` at all by passing this on the kernel command line! You could do the same with `modprobe`. Or if you still need `dmsetup`, you could use that.

```
dm-mod.create="root,,,ro,0 1884160 verity 1 PARTUUID=$boot_uuid/PARTNROFF=1 PARTUUID=$boot_uuid/PARTNROFF=2 4096 4096 235520 1 sha256 8b461e3a50dfdf97d78d8d52e90772bb44c9d3b376d7f662adaea98b8202e57e 808fcbd183fa45e9fb787fcc4732e39eb70db61e26d4b21ba6896717ebe9d4f9 2 restart_on_corruption ignore_zero_blocks"
```

This probably won't be enough if we decide to sign the verity hash, but I think that's off the table now.

minimal-init
```diff
@@ -0,0 +1,194 @@
...
+    true # Don't set "norecovery" when mounting rw
+  else
+    usrflags="${usrflags},rescue=nologreplay"
+  fi
+fi
+# Only proceed if the source is a path.
+if echo "${usr}" | grep -v -q "^/"; then
+  echo "No mountable /usr partition given (usr='${usr}')" >&2
+  false # Throw error
+fi
```
chewi:
```suggestion
case "${usr}" in
  /*) : ;;
  *) echo "No mountable /usr partition given (usr='${usr}')" >&2
    false # Throw error
    ;;
esac
```

chewi (changes requested):
Well, it's far more comprehensive than what I was shooting for. I guess I shouldn't assume that it will always just work!

Was there any particular reason why you went with squashfs over erofs in the end, other than mkfs.erofs not being in the SDK? It probably should be.

Even without these changes, the verity-generator module and much of the usr-generator module is redundant anyway following changing in systemd. I have already dropped that code in my sd-veritysetup branch, which was originally going to be part of my boot rework, but I was also planning to apply it regardless. Generally speaking, I don't like leaving dead code around. That's what git is for. :wink:

I recall someone saying something about our existing boot process being susceptible to duplicate disks, where it might pick USR-A from a disk other than the one you actually booted from. Maybe we could address that here (later?) although I'm not sure how you would tell which was the boot disk.

minimal-init
pothos: I think there's a config option for busybox that selects how built-in execution is done. Anyway, the --install -s line sets up the symlinks and afterwards everything is prepared to also be reachable through the PATH.

minimal-init
pothos: The modules.alias file maps these to module names, if a module exists for it.

minimal-init
pothos: Yeah, but one one adds it (or copies this line to another script), this breaks, so I think we better have this.

minimal-init
pothos: Thanks

pothos:
> Was there any particular reason why you went with squashfs over erofs in the end, other than mkfs.erofs not being in the SDK? It probably should be.

So far we use squashfs for everything and I kept using it here for consistency. If we do it, I would prefer a follow-up PR.

> Even without these changes, the verity-generator module and much of the usr-generator module is redundant anyway following changing in systemd. I have already dropped that code in my sd-veritysetup branch, which was originally going to be part of my boot rework, but I was also planning to apply it regardless. Generally speaking, I don't like leaving dead code around. That's what git is for. 😉

Yes, good point. The generators don't handle the missing `mount.` prefix, or - but that could also be gained by preprocessing? I kept the files around because I thought we might go back to this - either now when things don't work on real hardware or later if we have enough space again - but I'm also ok with removing entirely and relying on git.

chewi:
> So far we use squashfs for everything and I kept using it here for consistency. If we do it, I would prefer a follow-up PR.

Okay. I thought we were using erofs in the bakery now, which we are, but it's still not the default.

> The generators don't handle the missing `mount.` prefix, or - but that could also be gained by preprocessing?

Sorry, I don't know what you're referring to there.

minimal-init
chewi: busybox doesn't support it, and I would say that if anyone copies this code for use outside busybox, then they're probably doing something bad. :stuck_out_tongue: Your call!

pothos:
The systemd generator uses `mount.usr=` but the generators we have are for `usr=` which apparently was used before - I didn't check how old the nodes would be and if we can drop support for this. Probably ok if this the switch was already done pre-Flatcar?

chewi:
Ah! See [this wrapper](https://github.com/flatcar/bootengine/commit/4d3ce13b241196a95c334333e81376749f721e0f). This is academic in the case of the verity-generator though since you're now handling it here. I just meant that it's redundant either way.

minimal-init
pothos: A recent enough busybox supports it now. I haven't set `-eo pipefail` up here though because of the emergency shell trap that we have.

pothos:
I've removed the unused code now and agree that we can probably make use of systemd upstream stuff in the future.
The support for `usr=` instead of `mount.usr` is still kept, though and when we remove it then I would rather want to do it as a follow-up.

minimal-init
chewi: On reflection, I guess `veritysetup` isn't a heavy requirement.

chewi (approved):

minimal-init
pothos: With the libs it pulls in it is considerable but since the overall size wins are big enough I accepted this cost. Interesting that these dm-mod parameters are also possible!

minimal-init
```diff
@@ -0,0 +1,196 @@
...
+mkdir /work/realinit /work/work
+mount -t squashfs "${LOOP}" /underlay
+mkdir -p /realinit
+mount -t overlay -o rw,lowerdir=/underlay,upperdir=/work/realinit,workdir=/work/work overlay /realinit
+mkdir -p /realinit/sysusr/usr
+mount -o move /sysusr/usr /realinit/sysusr/usr
+if [ "${usr}" = /usr.squashfs ]; then
+  mkdir -p /oem
+  mkdir -p /realinit/oem
+  mount -o bind /oem /realinit/oem
```
chewi, line 185: @pothos, do you remember why you used `/oem` rather than `/usr/share/oem` here? Our stock initrd doesn't have the former, and the latter would have avoided flatcar/Flatcar#2023. With further changes, I could reduce our dependence on the legacy `/usr/share/oem` path, but I feel like I'm missing something here.

pothos: I think this topic came also up when we moved from `/usr/share/oem` to `/oem` because this also broke there?

I think we should do a check here for the path contents of both `/oem` and `/usr/share/oem` and use the first one that has something.

pothos: If the bind mount is the problem we could also try a `mv` to be sure that we don't use more RAM than needed?

Needs testing with the pxe qemu script again.

chewi: I think I'll just move `/usr/share/oem/*` to `/oem` first or something like that.  The other references to `/usr/share/oem` in bootengine can be changed to `/oem`. I don't think the bind mount is an issue, but I'll check.

PR #115, opened 2025-12-10
initrd-setup-root: Use systemd-confext instead of custom overlay mount

pothos:
So far we had a custom overlay mount for /etc that provided the A/B
    updated files from /usr in a lowerdir. Since then we upstreamed a
    mutable mode for sysext and confext.
    We can now switch over to it and provide a default confext by using the
    mutable mode. Because there is no atomic remount yet and also because
    we want to avoid daemon reloads during boot, this relies on a new skip
    logic in systemd-sysext/confext to only refresh in the final system boot
    when changes are found. Through only using verity images we know that no
    changes can be there because they get compared by hash and not mtime.
    When we would hit a refresh during boot then /etc contents are shortly
    gone and services sometimes fail during boot.
    A bit specific to Flatcar/Ignition is that we load confext twice in the
    initrd, once because we have a default confext that provides /etc
    contents for Ignition and a second time for loading user confexts for
    the final system (when users added new ones through Ignition).

My upstream PR is merged so that SYSTEMD_IN_INITRD=0 is not needed in the future but setting this here spares us one more backport patch.

## How to use

With the scripts PR that uses it

## Testing done

See scripts PR


chewi (approved):
I'm not very familiar with this, but it looks good. Please could you link the relevant unreleased systemd changes in this pull request for reference.

pothos:
> I'm not very familiar with this, but it looks good. Please could you link the relevant unreleased systemd changes in this pull request for reference.

Thanks, yes, I wanted to capture this somewhere and have now added it to the commit message in the scripts repo that ships the patches.

pothos:
Just noticed that we should not fail the boot when sysexts can't get merged (broken sysexts) but continue booting. Otherwise we would force people to use the serial/VGA console instead of SSH. For confext we can't really continue because then we lack /etc contents and failing properly is best - the errors one gets due to missing /etc files are very strange.

I've also added detecting when the service is masked and then don't merge sysexts in the initrd.

PR #116, opened 2026-01-08
dracut: move the module ordering to the 50-59 range (rebased)

chewi:
# dracut: move the module ordering to the 50-59 range

This is necessary for upgrading to Dracut 109. See #109 where this was originally submitted. I had to rebase the changes, and the contributor's fork has since been deleted.

Jozzsi didn't sign off their commit, but I've signed it off on their behalf, which I believe is okay. Their changes are purely renames anyway.

## How to use

Check that the system boots and compare the initrd contents with a recent nightly.

## Testing done

I've taken this change alongside a bump to Dracut 109. A [Jenkins run](http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/7231/cldsv/) has passed aside from known issues. There is obviously a lot of churn in the changes report from files moving around, but changes to the initrd itself are very small. None of the changes are unexplained.


ader1990 (approved):
LGTM, I think that we need to make sure that the most complex scenarios are passing: LUKS2 with tpm/tang, pxe and maybe disk related ones - RAID ones. As far as I know, the CI has those scenarios - althought I am not sure that the more complex ones in the docs are tested: https://www.flatcar.org/docs/latest/setup/security/luks/

chewi:
I've tested the second example in that doc with systemd-cryptenroll against the TPM. It worked fine. As discussed, I also compared the non-binary initrd contents against a recent nightly.  There are changes, but just because of the 109 bump, nothing really concerning.

I'll merge this now, but if you still want to do additional testing, you have time before this gets picked up along with the bump in the scripts repo.

PR #117, opened 2026-01-28
afterburn: sync network kargs with supported provider

tormath1:
There is now a support for ProxmoxVE and Kubevirt: https://github.com/coreos/afterburn/blob/8f6d5877c96b4d974f18a4782e6e275efdbd94d6/src/initrd/mod.rs#L20-L24


chewi:
I was going to pull in the network kargs sync into my own PR, but the support for KubeVirt hasn't actually been released in Afterburn yet. I'll take just the Proxmox change.

PR #118, opened 2026-03-03
Allow use of legacy OEM path while migrating modules to the new path

chewi:
# Allow legacy OEM path while migrating to new path

The new minimal-init passed the contents of `/oem` to the second stage, but not the legacy `/usr/share/oem` path that is still [documented](https://www.flatcar.org/docs/latest/installing/bare-metal/booting-with-pxe/#adding-a-custom-oem). This change moves either the former or the latter into the second stage. This is no longer done via a bind mount, as that adds unnecessary complexity.

At the same time, Dracut modules in the second stage can simply reference `/oem` rather than `/usr/share/oem` now that the first stage deals with the migration.

This addresses flatcar/Flatcar#2023.

## How to use

Try the [documented](https://www.flatcar.org/docs/latest/installing/bare-metal/booting-with-pxe/#adding-a-custom-oem) method. You can also try it with `/oem` instead of `/usr/share/oem`.

## Testing done

I've successfully done the above with both paths. I've also kicked off a [Jenkins run](http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/7513/cldsv/), which passed aside from bpf.execsnoop.

### DIFF
--- a/minimal-init
+++ b/minimal-init
@@ -192,8 +192,7 @@ mount -o move /sysusr/usr /realinit/sysusr/usr
 if [ "${usr}" = /usr.squashfs ]; then
   # Move either /oem (preferred) or /usr/share/oem (legacy) into the new root.
   mv /oem /realinit || mv /usr/share/oem /realinit || true
-  touch /realinit/usr.squashfs
-  mount -o bind /usr.squashfs /realinit/usr.squashfs
+  mv /usr.squashfs /realinit
 fi
 debug_sh 4/4: before switch_root to /realinit
 killall mdev || true
--- a/minimal-init
+++ b/minimal-init
@@ -190,9 +190,8 @@ mount -t overlay -o rw,lowerdir=/underlay,upperdir=/work/realinit,workdir=/work/
 mkdir -p /realinit/sysusr/usr
 mount -o move /sysusr/usr /realinit/sysusr/usr
 if [ "${usr}" = /usr.squashfs ]; then
-  mkdir -p /oem
-  mkdir -p /realinit/oem
-  mount -o bind /oem /realinit/oem
+  # Move either /oem (preferred) or /usr/share/oem (legacy) into the new root.
+  mv /oem /realinit || mv /usr/share/oem /realinit || true
   touch /realinit/usr.squashfs
   mount -o bind /usr.squashfs /realinit/usr.squashfs
 fi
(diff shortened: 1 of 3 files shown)


chewi:
CC @pothos

minimal-init
```diff
@@ -190,9 +190,8 @@ mount -t overlay -o rw,lowerdir=/underlay,upperdir=/work/realinit,workdir=/work/
 mkdir -p /realinit/sysusr/usr
 mount -o move /sysusr/usr /realinit/sysusr/usr
 if [ "${usr}" = /usr.squashfs ]; then
-  mkdir -p /oem
-  mkdir -p /realinit/oem
-  mount -o bind /oem /realinit/oem
+  # Move either /oem (preferred) or /usr/share/oem (legacy) into the new root.
+  mv /oem /realinit || mv /usr/share/oem /realinit || true
   touch /realinit/usr.squashfs
   mount -o bind /usr.squashfs /realinit/usr.squashfs
```
pothos: Would be good to do the move here, too, to keep it consistent.

chewi: Good idea.

chewi: Retested with that. All good.

minimal-init
```diff
@@ -190,11 +190,9 @@ mount -t overlay -o rw,lowerdir=/underlay,upperdir=/work/realinit,workdir=/work/
...
 mount -o move /sysusr/usr /realinit/sysusr/usr
 if [ "${usr}" = /usr.squashfs ]; then
-  mkdir -p /oem
-  mkdir -p /realinit/oem
-  mount -o bind /oem /realinit/oem
-  touch /realinit/usr.squashfs
-  mount -o bind /usr.squashfs /realinit/usr.squashfs
+  # Move either /oem (preferred) or /usr/share/oem (legacy) into the new root.
+  mv /oem /realinit || mv /usr/share/oem /realinit || true
+  mv /usr.squashfs /realinit
```
krnowak, line 195: Excuse my lack of understanding here, but shouldn't it try to bind mount `/oem` (or `/usr/share/oem` as a fallback) into `/realinit/oem`?

pothos (approved):

krnowak (commented):
Gah. I thought I submitted the review, but apparently I did not…

minimal-init
chewi: Bind mounting /oem is what it did before. There was no need to though because this is all just initrd data that's otherwise about to get thrown away. Might as well just move it to where we need it.

krnowak: Thanks.

PR #121, opened 2026-04-15
ignition: Mount OEM partition so that Ignition doesn't have to do it

chewi:
# Mount OEM partition so that Ignition doesn't have to do it

Ignition no longer does its own mounting of the OEM partition for the `oem://` scheme, so bootengine needs to mount it in advance. The `oem://` scheme therefore now maps to `/realinit/oem`, but Ignition may also write to the same partition at `/realinit/sysroot/oem`.

The hope is to eventually drop even more patches from Ignition and use a new `file://` scheme with Butane doing the mapping from `oem://` to `file:///oem`.

Rather than shunting OEM data between `/oem`, `/realinit/oem`, and `/realinit/sysroot/oem`, mount `/realinit/oem` as a tmpfs from the start (for PXE) and turn `/realinit/sysroot/oem` into a bind mount.

We want to unmount the disk-based OEM partition before Ignition's disk stage, in case it gets wiped, and its mount stage, in case it gets remounted. This can be done more simply with `findmnt`.

However, we do want the disk-based OEM partition to be mounted while Ignition's fetch-offline and fetch stages run in case the user has put config there. This is now done with `RequiresMountsFor` now that Ignition no longer does it for us. The kargs stage can also use this mount to write `grub.cfg` rather than mounting it again in its own namespace.

The `nodev` option wasn't applied to PXE.

We only expect a disk-based OEM partition to be btrfs or ext4, so don't rely on the `auto` type.

Aside from these mount-related changes, the ignition-setup script has been dropped (also by upstream) as it is no longer needed, I have reworked the Ignition units to make them easier to compare against upstream, and I have changed the `emergency.target` job mode from `isolate` to `replace-irreversibly`.

The `isolate` job mode causes `dev-mapper-usr.device` and therefore `sysusr-usr.mount` to stop. We cannot stop the latter though because the emergency shell is hosted within it. It's not entirely clear why this happens because both units have `IgnoreOnIsolate=yes`. Trying to unmount doesn't break the shell, but it does show an unexpected error.

The usual approach of preventing the above with drop-in units doesn't seem to work for auto-generated mount units. However, I noticed that systemd itself has not used the `isolate` job mode for `emergency.target` since 2013. `replace-irreversibly` does not proactively stop all the other units, just the ones that would block the new target. This can actually make debugging easier in the event of an Ignition failure because the system will be in a closer state to the one when the failure occurred.

See flatcar/scripts#3934 for the associated Ignition changes (in scripts).

## How to use

You can try variations on these configs:

```yaml
variant: flatcar
version: 1.0.0
storage:
  filesystems:
    - path: /oem
      device: /dev/disk/by-label/OEM
      format: ext4
      wipe_filesystem: true
      label: OEM
      with_mount_unit: true
  files:
    - path: /oem/foo
      overwrite: true
      contents:
        inline: |
          bar
```

```yaml
variant: flatcar
version: 1.0.0
storage:
  files:
    - path: /etc/baz
      overwrite: true
      contents:
        source: oem:///mybaz.txt
```

For the latter to work, you have to place mybaz.txt in the OEM partition of the image beforehand (use `guestmount`) or add it to an initrd alongside the config and append that initrd to the PXE image.

Note that Butane does not currently support `oem://` so you actually need to put `gs://` initially and then change it to `oem://` in the resulting JSON.

You can also try loading alternative configs from the OEM partition by passing `ignition.config.url=oem:///myconfig.ign` on the kernel command line.

## Testing done

[Jenkins has passed](https://jenkins.flatcar.org/job/container/job/packages_all_arches/38/cldsv/) with these changes alongside the associated Ignition changes. I have also done lots of manual testing.


krnowak:
Is wiping OEM partition in ignition essentially a noop, since some tmpfs will be wiped instead and then, at a later stage, we mount the disk-based OEM partition?

chewi:
The tmpfs OEM is for PXE only where there is no Flatcar disk (yet). It wouldn't make sense to try and wipe it in that case. I don't know why anyone would want to wipe it even in the non-PXE case, but apparently that is supposed to work.

dracut/51diskless-generator/diskless-generator
```diff
@@ -57,6 +54,24 @@ What=/usr.squashfs
...
+    # When /usr is provided by the initrd, then OEM is a tmpfs. This is already
+    # mounted by minimal-init, so we don't need a unit in that case.
+    cat >"${UNIT_DIR}/oem.mount" <<EOF
+# Automatically generated by diskless-generator
+
+[Unit]
+# This must be mounted before local-fs-pre.target.
+DefaultDependencies=no
+Before=umount.target
+Conflicts=umount.target
```
krnowak: The comment is confusing - "must be mounted before local-fs-pre.target", but you specify `Before=umount.target`?

chewi: The `umount.target` lines are added by default if you don't specify `DefaultDependencies=no`. They're used during shutdown.

krnowak: Yeah, I just can't find a connection between the comment and the lines that follow it.

chewi: Okay, fixed.

krnowak (approved):

PR #124, opened 2026-04-30
ignition: Improve OEM detection and fix VirtualBox

chewi:
# ignition: Improve OEM detection and fix VirtualBox

This falls back to detection for cases unsupported by Ignition, which saves us from having to patch these cases into Ignition like we have previously been doing.

It also means we can make a more useful guess than just "metal", which does nothing. For example, this allows you to pass config via qemu_fw when testing PXE with QEMU.

The vboxguest kernel module is added to support VirtualBox against new Ignition versions.

This also drops Equinix Metal (Packet) support.

## How to use

The easiest image to test this with is Vagrant... which isn't that easy. You'll need my wider OEM changes.

## Testing done

I've tried the above manually run [Jenkins](https://jenkins.flatcar.org/job/container/job/packages_all_arches/120/cldsv/), which passed where expected (eventually).

### DIFF
--- a/dracut/53ignition/ignition-setup-pre.sh
+++ b/dracut/53ignition/ignition-setup-pre.sh
@@ -3,7 +3,7 @@
 
 set -e
 
-cmdline=( $(</proc/cmdline) )
+read -r -a cmdline < /proc/cmdline
 cmdline_arg() {
     local name="$1" value="$2"
     for arg in "${cmdline[@]}"; do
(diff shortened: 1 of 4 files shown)


dracut/53ignition/ignition-setup-pre.sh
```diff
@@ -15,28 +15,28 @@ cmdline_arg() {
...
-# Ignition changed the platform name to "aws"
-if [ "${oem_cmdline}" = "ec2" ]; then
-  oem_cmdline="aws"
-fi
-
-# Ignition changed the platform name to "gcp"
-if [ "${oem_cmdline}" = "gce" ]; then
-  oem_cmdline="gcp"
-fi
+case $(systemd-detect-virt) in
```
krnowak, line 19: I see that `case` swallows failures (`systemd-detect-virt` returns 1 when there is no virtualization). This `set -e` feature seems to be of questionable usefulness.

krnowak (approved):

dracut/53ignition/ignition-setup-pre.sh
chewi: Well, that is what I want in this case!

PR #125, opened 2026-05-14
switch-root: Ensure sysctl is run again after switching root

chewi:
Dracut normally copies all sysctl config to the initrd, but we may want to exclude platform-specific config because the initrd is shared between platforms. That config can be picked up from the OEM sysext after switching root.

I have tested this manually with GCE.


krnowak (approved):
I assume that this is for picking up sysctl files that are distributed with an OEM sysext, right? Is systemd-sysctl service ordered to run after systemd-{sys,conf}ext?

PR #129, opened 2026-08-26
flatcar-network: Use upstream network and resolved Dracut modules

chewi:
Our IP parser has some shortcomings and systemd-network-generator can simply do this for us. Afterburn is being modified (coreos/afterburn#1299) to emit an environment file with `SYSTEMD_PROC_CMDLINE` because the generator won't read the existing /etc/cmdline.d file that was consumed by Dracut.

This also deduplicates .network files in common with coreos-init. The bootengine copies had gone out of sync, and we can just install coreos-init's copies. We need to adjust `KeepConfiguration` for the initrd environment though.

This partially addresses #117 by allowing afterburn-network-kargs.service to work with ProxmoxVE. We cannot allow KubeVirt yet because a new Afterburn release is needed first.

The rest is just small cleanups.

## How to use

This is most effectively tested with ProxmoxVE. /run/systemd/networkd should include 20-eth0.network and 71-default.network generated from the two different mechanisms. They should be roughly similar.

## Testing done

The most recent [Jenkins build](https://jenkins.flatcar.org/job/container/job/packages_all_arches/575/cldsv/) was successful. Earlier builds tested all the platforms successfully. I have also done a lot of manual testing with QEMU and ProxmoxVE.


dracut/50flatcar-network/module-setup.sh
```diff
@@ -16,7 +16,16 @@ install() {
         /etc/systemd/resolved.conf
 
     inst_multiple \
-        "$systemdnetwork"/{20-calico-tunl0,50-veth,98-{gce-coreos-virtio,gce-virtio,virtio}}.link
+        "$systemdnetwork"/{20-calico-tunl0,50-veth,98-{gce-coreos-virtio,gce-virtio,virtio}}.link \
+        "$systemdnetwork"/yy-{azure-sriov{,-coreos},pxe}.network \
+        "$systemdnetwork"/zz-default.network
```
tormath1: If those are provided by `coreos-init` - should we add it as a post dependency to the bootengine ebuild?

tormath1 (approved):
One question - but it looks good to me. It simplifies a lot this area. Thanks for the deep dive!

dracut/50flatcar-network/module-setup.sh
chewi: It would make sense given that `update-bootengine` chooses the modules, but we've always put these dependencies in coreos-kernel for some reason. I'll stick with that for now. We can do a big cleanup later, although I'm also still thinking about dropping Dracut entirely.

